PR #26, opened 2015-10-01
Declarative API for specifying nested fields

taion:


taion:
cc @cancan101

flask_jsonapiview/view.py
```diff
@@ -108,6 +108,8 @@ class ModelView(ApiView):
     model = None
     url_id_key = 'id'
 
+    nested = None
```
cancan101: Make this `{}`.

cancan101: or frozen dict.

taion: That's mutable though...

cancan101: http://werkzeug.pocoo.org/docs/0.10/datastructures/#werkzeug.datastructures.ImmutableDict

taion: Think it's worth keeping the check at https://github.com/4Catalyzer/flask-jsonapiview/pull/26/files#diff-7e2c0e4609fe4eecb154d2d2534b07d4R183, then?

PR #68, opened 2015-11-29
Do not use flask.ext

taion:


taion:
Per https://github.com/mitsuhiko/flask/issues/1092.

PR #75, opened 2015-12-28
add apispec plugin

itajaja:
with this pr is possible to use `flask_res.spec` as a plugin to [apispec](https://github.com/marshmallow-code/apispec) to generate swagger options.

Some remarks: I don’t know if the idea of subclassing `SpecDeclaration`
is a good way to do it. I tried to automate as many things as possible,
with the result that the end result might not always be correct. The
View can always subclass SpecDeclaration and correct any errors coming
from the view superclasses.

### DIFF
--- a/flask_resty/api.py
+++ b/flask_resty/api.py
@@ -19,7 +19,8 @@ class Api(object):
         self._prefix = prefix
 
     def init_app(self, app):
-        app.extensions['resty'] = self
+
+        app.extensions['resty'] = FlaskRestyState(self)
 
         @app.errorhandler(ApiError)
         def handle_api_error(error):
@@ -28,6 +29,8 @@ class Api(object):
     def _get_app(self, app):
         app = app or self._app
         assert app, "no application specified"
+        # store the view urls for reference. Doesn't support multiple routes
+        # mapped to same view
         return app
 
     def add_resource(
@@ -44,6 +47,7 @@ class Api(object):
             assert id_rule is None
 
         app = self._get_app(app)
+        views = app.extensions['resty'].views
         endpoint = self._get_endpoint(base_view, alternate_view)
 
         base_rule_full = '{}{}'.format(self._prefix, base_rule)
@@ -51,6 +55,7 @@ class Api(object):
 
         if not alternate_view:
             app.add_url_rule(base_rule_full, view_func=base_view_func)
+            views[base_view] = Resource(base_view, base_rule_full)
             return
 
         alternate_rule_full = '{}{}'.format(self._prefix, alternate_rule)
@@ -71,6 +76,9 @@ class Api(object):
             methods=alternate_view.methods,
         )
 
+        views[base_view] = Resource(base_view, base_rule_full)
+        views[alternate_view] = Resource(alternate_view, alternate_rule_full)
+
     def _get_endpoint(self, base_view, alternate_view):
         base_view_name = base_view.__name__
         if not alternate_view:
@@ -89,3 +97,17 @@ class Api(object):
         @app.route(rule)
         def ping():
             return '', 200
+
+
+class FlaskRestyState(object):
+    def __init__(self, api):
+        self.api = api
+        self.views = {}
+
+
+class Resource(object):
+    """simple object to store information about an added resource"""
+
+    def __init__(self, view, rule):
+        self.rule = rule
+        self.view = view
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -5,9 +5,10 @@ from marshmallow import ValidationError
 import sqlalchemy as sa
 from sqlalchemy import Column, sql
 
-from .exceptions import ApiError
 from . import meta
 from . import utils
+from .exceptions import ApiError
+
 
 # -----------------------------------------------------------------------------
 
@@ -63,6 +64,18 @@ class LimitPagination(object):
 
         return limit
 
+    def spec_declaration(self, path, spec, **kwargs):
+        path['get'].add_parameter(name='limit',
+                                  type='int',
+                                  description='pagination limit')
+        path['get'].add_property_to_response(
+            prop_name='meta',
+            type='object',
+            properties={
+                'has_next_page': {'type': 'boolean'}
+            }
+        )
+
 
 class LimitOffsetPagination(LimitPagination):
     offset_arg = 'offset'
@@ -91,6 +104,13 @@ class LimitOffsetPagination(LimitPagination):
             raise ApiError(400, {'code': 'invalid_offset'})
 
         return offset
+
+    def spec_declaration(self, path, spec, **kwargs):
+        super(LimitOffsetPagination, self).spec_declaration(path, spec)
+        path['get'].add_parameter(
+            name='offset',
+            type='int',
+            description='pagination offset')
 
 
 class PagePagination(LimitOffsetPagination):
@@ -125,6 +145,13 @@ class PagePagination(LimitOffsetPagination):
 
     def get_limit(self):
         return self._page_size
+
+    def spec_declaration(self, path, spec, **kwargs):
+        super(PagePagination, self).spec_declaration(path, spec)
+        path['get'].add_parameter(
+            name='page',
+            type='int',
+            description='page number')
 
 
 class IdCursorPagination(LimitPagination):
@@ -265,3 +292,10 @@ class IdCursorPagination(LimitPagination):
         cursor = cursor.encode('utf-8')
         cursor = base64.urlsafe_b64encode(cursor)
         return cursor.decode('ascii')
+
+    def spec_declaration(self, path, spec, **kwargs):
+        super(IdCursorPagination, self).spec_declaration(path, spec)
+        path['get'].add_parameter(
+            name='cursor',
+            type='string',
+            description='pagination cursor')
--- a/flask_resty/sorting.py
+++ b/flask_resty/sorting.py
@@ -71,3 +71,9 @@ class Sorting(SortingBase):
             })
 
         return super(Sorting, self).get_column(view, field_name)
+
+    def spec_declaration(self, path, spec, **kwargs):
+        path['get'].add_parameter(
+            name='sort',
+            type='string',
+            description='field to sort by')
--- a/flask_resty/spec/__init__.py
+++ b/flask_resty/spec/__init__.py
@@ -1,0 +1,9 @@
+# flake8: noqa
+
+"""Flask-Resty plugin for apispec. Allows passing a Flask-Resty
+`view` to `APISpec.add_path <apispec.APISpec.add_path>`
+"""
+
+from .plugin import setup, schema_path_helper
+from .declaration import *
+from .utils import ref, flask_path_to_swagger, get_marshmallow_schema_name
--- a/flask_resty/spec/declaration.py
+++ b/flask_resty/spec/declaration.py
@@ -1,0 +1,94 @@
+from flask.views import http_method_funcs
+
+from .utils import get_marshmallow_schema_name, ref
+
+
+class ApiViewDeclaration(object):
+    """Simple Declaration for ApiView Classes
+    :param many: whether or not this view describes a list or a single instance
+    :param tag: whether or not the schema name should be added as a tag"""
+
+    def __init__(self, many=False, tag=True, **kwargs):
+        self.many = many
+        self.tag = tag
+
+        invalid_kwargs = list(set(kwargs.keys()) - http_method_funcs)
+        if invalid_kwargs:
+            raise TypeError(
+                'invalid keyword argument "{}"'.format(invalid_kwargs[0]))
+
+        self.overrides = kwargs
+
+    def __call__(self, view, path, spec):
+        view_methods = set(view.__dict__.keys()) \
+            .intersection(http_method_funcs)
+
+        for method in view_methods:
+            operation = path[method]  # triggers the operation generation
+            if getattr(view, method).__doc__:
+                # add docstring as description
+                operation['description'] = getattr(view, method).__doc__
+
+        if view.schema:
+            self.declare_schema(view, view_methods, path, spec)
+
+        for method in view_methods.intersection(self.overrides.keys()):
+            for code, response in self.overrides[method].items():
+                path[method].declare_response(code, **response)
+
+    def declare_schema(self, view, view_methods, path, spec):
+        schema = get_marshmallow_schema_name(spec, type(view.schema))
+        schema_ref = ref(schema)
+
+        if 'get' in view_methods:
+            if not self.many:
+                data = schema_ref
+            else:
+                data = {
+                    'type': 'array',
+                    'items': schema_ref
+                }
+            path['get'].add_property_to_response(prop_name='data', **data)
+
+        if 'post' in view_methods:
+            path['post'].add_parameter('body',
+                                       name='body',
+                                       required=True,
+                                       schema=schema_ref)
+            path['post'].declare_response(201)
+
+        if 'put' in view_methods:
+            path['put'].add_parameter('body',
+                                      name='body',
+                                      required=True,
+                                      schema=schema_ref)
+            path['put'].declare_response(204)
+
+        if 'patch' in view_methods:
+            path['patch'].add_parameter('body',
+                                        name='body',
+                                        required=True,
+                                        schema=schema_ref)
+            path['patch'].add_property_to_response(prop_name='data',
+                                                   **schema_ref)
+
+        if 'delete' in view_methods:
+            path['delete'].declare_response(204)
+
+        if self.tag:
+            for method in view_methods:
+                path[method].add_tag(schema)
+
+
+class ModelViewDeclaration(ApiViewDeclaration):
+    """Declaration for Views that specify a model"""
+
+    def __call__(self, view, path, spec):
+        super(ModelViewDeclaration, self)\
+            .__call__(view, path, spec)
+
+        for item in (view.pagination, view.filtering, view.sorting):
+            try:
+                item.spec_declaration(path, spec)
+            except AttributeError:
+                pass
--- a/flask_resty/spec/operation.py
+++ b/flask_resty/spec/operation.py
@@ -1,0 +1,44 @@
+class Operation(dict):
+    """Exposes some utility methods to compose a swagger operation
+    https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object
+    """
+
+    def __init__(self):
+        super(Operation, self).__init__({
+            'responses': {},
+            'parameters': []
+        })
+
+    def add_parameter(self, location='query', **kwargs):
+        """Adds a new parameter to the request
+        :param location: the 'in' field of the parameter (e.g: 'query',
+            'body', 'path')
+        """
+        kwargs.setdefault('in', location)
+        if kwargs['in'] != 'body':
+            kwargs.setdefault('type', 'string')
+        self['parameters'].append(kwargs)
+        pass
+
+    def add_property_to_response(self, code='200', prop_name='data', **kwargs):
+        """Add a property (http://json-schema.org/latest/json-schema-validation.html#anchor64) # noqa
+        to the schema of the response identified by the code"""
+        self['responses']\
+            .setdefault(str(code), self._new_operation())\
+            .setdefault('schema', {'type': 'object'})\
+            .setdefault('properties', {})\
+            .setdefault(prop_name, {})\
+            .update(**kwargs)
+
+    def declare_response(self, code='200', **kwargs):
+        """Declare a response for the specified code
+        https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#responseObject # noqa"""
+        self['responses'][str(code)] = self._new_operation(**kwargs)
+
+    def add_tag(self, tag):
+        """Add a tag to the operation"""
+        self.setdefault('tags', []).append(tag)
+
+    def _new_operation(self, **kwargs):
+        kwargs.setdefault('description', '')
+        return dict(kwargs)
--- a/flask_resty/spec/plugin.py
+++ b/flask_resty/spec/plugin.py
@@ -1,0 +1,45 @@
+from collections import defaultdict
+
+from flask import current_app
+
+from .operation import Operation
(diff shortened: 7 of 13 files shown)


flask_resty/api.py
```diff
@@ -16,6 +16,10 @@ def __init__(self, app=None, prefix=''):
             self._app = app
             self.init_app(app)
 
+        # store the view urls for reference. Doesn't support multiple routes
+        # mapped to same view
+        self.views = {}
```
taion: I don't think this is safe - plugins should support working w/multiple apps (e.g. see `init_app` below).

itajaja: Uhm, I really need that additional information during `_add_resource`... I don't know what's the best place to store it though.... Creating another dictionary in `app.extensions`? :worried:

taion: Can you just make it a dict of `app` -> `views`?

itajaja: Have to check if I have the reference to the app in the location I have to use it - but I think so

flask_resty/api.py
```diff
@@ -28,6 +27,12 @@ def handle_api_error(error):
     def _get_app(self, app):
         app = app or self._app
         assert app, "no application specified"
+        # store the view urls for reference. Doesn't support multiple routes
+        # mapped to same view
+        app.extensions.setdefault('resty', {
```
taion: This should be an actual class. See e.g. https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L819.

flask_resty/pagination.py
```diff
@@ -5,14 +5,16 @@
 import sqlalchemy as sa
 from sqlalchemy import Column, sql
 
+from .spec import SpecDeclaration
```
taion: nit: alphabetize

flask_resty/api.py
itajaja: ok, what about

```
app.extensions.setdefault('resty', _FlaskRestyState(self, views={})
```

flask_resty/pagination.py
```diff
@@ -63,6 +65,17 @@ def parse_limit(self, limit):
 
         return limit
 
+    def update_spec(self, path, spec, **kwargs):
+        path['get'].add_parameter(name='limit',
+                                  type='int',
+                                  description='pagination limit')
+        path['get'].add_property_to_response(prop_name='meta', **{
```
taion: Why do you need the `**` here?

taion: I mean, instead of just using normal kwargs.

flask_resty/pagination.py
```diff
@@ -265,3 +292,10 @@ def encode_cursor(self, cursor):
         cursor = cursor.encode('utf-8')
         cursor = base64.urlsafe_b64encode(cursor)
         return cursor.decode('ascii')
+
+    def update_spec(self, path, spec, **kwargs):
+        super(IdCursorPagination, self).update_spec(path, spec)
+        path['get'].add_parameter(
+            name='cursor',
+            type='int',
```
taion: `cursor` is an opaque string, not an int.

flask_resty/sorting.py
```diff
@@ -49,7 +50,7 @@ def __call__(self, query, view):
         return self.sort_query(query, view, self._field_specs)
 
 
-class Sorting(SortingBase):
+class Sorting(SortingBase, SpecDeclaration):
```
taion: Possibly `SortingBase` should extend `SpecDeclaration`? And just raise `NotImplementedError()` (or do nothing).

flask_resty/spec/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Flask-Resty plugin for apispec. Allows passing a Flask-Resty
+`view` to `APISpec.add_path <apispec.APISpec.add_path>`
+"""
+
+from .plugin import setup, schema_path_helper  # noqa
+from .swagger_utils import *  # noqa
```
taion: nit: alpha

flask_resty/pagination.py
itajaja: no need. gonna change it to use named args

flask_resty/pagination.py
itajaja: :+1:

flask_resty/spec/operation.py
```diff
@@ -0,0 +1,43 @@
...
+        :param location: the 'in' field of the parameter (e.g: 'query',
+            'body', 'path')
+        """
+        kwargs.setdefault('in', location)
+        if kwargs['in'] != 'body':
+            kwargs.setdefault('type', 'string')
+        self['parameters'].append(kwargs)
+        pass
+
+    def add_property_to_response(self, code='200', prop_name='data', **kwargs):
```
taion: What's up with stringifying the code here? Why not just leave it as an int?

flask_resty/sorting.py
itajaja: Yeah, i guess it's cleaner. (with `pass` though, not raising the exception)

flask_resty/spec/plugin.py
```diff
@@ -0,0 +1,45 @@
+from collections import defaultdict
+
+from flask import current_app
+from flask.ext.resty.spec.operation import Operation
+from .swagger_utils import flaskpath2swagger
+
+RESTY_PLUGIN_NAME = 'resty'
```
taion: Pull this out to something like https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L608-L613 (maybe in a separate top-level `.utils`).

flask_resty/spec/plugin.py
```diff
@@ -0,0 +1,45 @@
+from collections import defaultdict
+
+from flask import current_app
+from flask.ext.resty.spec.operation import Operation
```
taion: `from .operation import ...`

All the imports in here should be relative.

flask_resty/spec/spec_declaration.py
```diff
@@ -0,0 +1,9 @@
+class SpecDeclaration(object):
+
+    def update_spec(self, path, spec, **kwargs):
+        """When overridden, gives a chance to update the path fragment of the
+        specification
+        :param dict path: swagger-spec/blob/master/versions/2.0.md#pathItemObject) # noqa
+        :param apispec.APISpec spec: the specification object
+        """
+        pass
```
taion: What happens if you make this `raise NotImplementedError()`?

flask_resty/spec/swagger_utils.py
```diff
@@ -0,0 +1,31 @@
...
+RE_SWAGGER_URL = re.compile(r'<(?:[^:<>]+:)?([^<>]+)>')
+MARSHMALLOW_PLUGIN_NAME = 'apispec.ext.marshmallow'
+
+
+def ref(schema_name, path='definitions'):
+    """Generate a ref object"""
+    return {'$ref': '#/{}/{}'.format(path, schema_name)}
+
+
+def flaskpath2swagger(path):
```
taion: keep consistent fn naming - should be `flask_path_to_swagger` or equivalent

flask_resty/spec/__init__.py
taion: Also, can this just be `utils`? And the below just `declaration`? You're already making a separate package; no need to further namespace.

flask_resty/spec/operation.py
itajaja: swagger specs suggest to use strings instead of integers. Also to eliminate any possible difference arising from inserting 200 and "200"

setup.py
```diff
@@ -44,13 +44,14 @@ def run(self):
...
     keywords='rest flask',
-    packages=('flask_resty',),
+    packages=find_packages(exclude='tests'),
     install_requires=(
         'Flask >= 0.10',
         'Flask-SQLAlchemy >= 1.0',
         'marshmallow >= 2.2.0',
         'SQLAlchemy >= 1.0.0',
         'Werkzeug >= 0.11',
+        'apispec >= 0.3.2',
```
taion: Should this be under an `extra`?

flask_resty/spec/plugin.py
itajaja: :+1:

flask_resty/spec/spec_declaration.py
itajaja: I just find it safer. it's more of an "optional" interface

flask_resty/view.py
```diff
@@ -107,8 +108,62 @@ def validate_request_id(self, data, expected_id):
         if id != expected_id:
             raise ApiError(409, {'code': 'invalid_id.mismatch'})
 
+    def update_spec(self, path, spec, many=False, tag=False, **kwargs):
```
taion: Could we move this into a mixin, perhaps? Would give more composability w/r/t how to do this.

flask_resty/spec/swagger_utils.py
itajaja: I guess you are right. I blidly copied it from https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/flask.py#L31:)

flask_resty/api.py
taion: You'd put the `views={}` into the ctor, but otherwise yeah. No need to prepend with underscore; we're not exporting that class anyway.

setup.py
itajaja: Let me try if it works without the dependency.

flask_resty/view.py
itajaja: Can you expand on this?

flask_resty/spec/spec_declaration.py
taion: If it's optional, they just shouldn't implement it. If necessary we can then do `isinstance` checks or whatever.

flask_resty/view.py
taion: Literally just pull it out into a `ApiSpecViewMixin` or something. Then if they want this behavior, they can do

```py
class WidgetView(GenericModelView, SpecViewMixin)
```

flask_resty/spec/spec_declaration.py
itajaja: Still, The inheritance chain could be complicated, and users could not alway know if they are inheriting it or not from some abstract ancestor That doesn't implement it, or something like that. But you have a good point, I guess it's fine to implement it as `raise`

flask_resty/view.py
itajaja: This will require changing all the views in the services we are using though

taion: Isn't that a single base view almost everywhere?

itajaja: and since there is a different implementation for `ApiView` vs `GenericModelView` will the client have to decide which one to chose? Or the mixin will have an `isinstance` to differentiate?

itajaja: > Isn't that a single base view almost everywhere?

still, it requires to change all the services

flask_resty/spec/spec_declaration.py
```diff
@@ -0,0 +1,9 @@
+class SpecDeclaration(object):
+
+    def update_spec(self, path, spec, **kwargs):
```
taion: Can this be a class method?

flask_resty/spec/spec_declaration.py
taion: To elaborate - I think when we call into `update_spec`, we should consider using duck typing instead of doing an actual type check... probably more Pythonic that way.

flask_resty/spec/spec_declaration.py
itajaja: would I be able to do `cls.schema` on the views for example?

taion: Actually I guess not - since we do `self.whatever` everywhere, the relevant attributes might only be available on instances.

itajaja: yes. This has the unfortunate drawback that you actually have to initialize the view when you generate the documentation, which might have some undesired side effect

flask_resty/view.py
taion: Maybe `ApiSpecMixin` and `ModelApiSpecMixin`? It's just that this feels in some sense like "extra" behavior that should be opt-in.

flask_resty/spec/spec_declaration.py
taion: Nah, initializing the view should be fine - or at least side-effect free.

flask_resty/view.py
itajaja: but then the client could use `ModelApiSpecMixin` on Api View and vice versa. I understand what you are saying, but I don't know if mixins are the best way for this. My intent was actually to couple the `update_spec` to the actual object being specified. update_spec should change when the class is changed, so it makes sense to have them on the same file

taion: Maybe a better way to do this is to make it consistent with the other customization hooks. Do something like

```py
class WidgetView(GenericModelView):
    api_spec = ModelApiSpec()

    def get(self, id):
        return self.retrieve(id)
```

The idea here is that it'd be nice to separate out the inheritance hierarchy of the view spec generation from that of the views themselves - that way you could have better re-use of custom spec generation across multiple unrelated views as needed.

itajaja: a couple of questions with this.

- would you do the same for pagination, sorting, filtering?
- Does `ModelApiSpec` in your example return a function or the actual object? Because it needs a lot of parameters.

The idea of inheritance is that it's easy to call each superclass in order and each superclass has a chance of changing apispec, it's almost as update_spec is a more formal docstring. I'll experiment with what you are proposing, to see how the api turns out

flask_resty/spec/spec_declaration.py
itajaja: by duck typing you mean doing something like

````py
if callable(getattr(object, 'update_spec'))
````

As I said in another, I find inheritance very handy to express the update_spec mechanism. Not sure how easy would be without inheritance.

Comment (taion) [inline flask_resty/spec/spec_declaration.py]: `if hasattr(obj, 'update_spec')` maybe... not sure. The inheritance is fine, but it's usually preferable to do duck typing over class checking in Python, especially in this case where you're just looking for a method to call.

**Comment by itajaja** [comment]
@taion I update the spec declaration mechanism, have a look at https://github.com/itajaja/flask-resty/commit/a5b66eac8651dd9a7bef021910d2e5fbe35e0c5d. There are a couple of things I am perplexed about:

- The view is bound to the specification method in https://github.com/4Catalyzer/flask-resty/commit/a5b66eac8651dd9a7bef021910d2e5fbe35e0c5d#diff-900ad3ef6f69b8cca4a8457ced41bb41R22, which sounds weird to me.
- The way the declaration can be extended (see https://github.com/4Catalyzer/flask-resty/commit/a5b66eac8651dd9a7bef021910d2e5fbe35e0c5d#diff-b06079ef87e70635d8caefa4a04c9c62R72) is not really elegant. I could have subclassed `ModelViewDeclaration`, but that would have put a `BarViewDeclaration` class that is really not used by anything else.

Also there is [something going on](https://coveralls.io/files/1822077566#L23) with the coverage, and I am not sure how my last commit could have broken that

**Comment by taion** [comment]
The coverage miss is here: https://coveralls.io/builds/4729593/source?filename=flask_resty%2Fspec%2Fdeclaration.py

I'll look at the rest in a bit.

File: flask_resty/api.py [hunk 16]
```diff
@@ -28,6 +27,9 @@ def handle_api_error(error):
     def _get_app(self, app):
         app = app or self._app
         assert app, "no application specified"
+        # store the view urls for reference. Doesn't support multiple routes
+        # mapped to same view
+        app.extensions.setdefault('resty', FlaskRestyState(self))
```
taion: Why not do this in `init_app`?

flask_resty/api.py
```diff
@@ -89,3 +96,18 @@ def add_ping(self, rule, app=None):
         @app.route(rule)
         def ping():
             return '', 200
+
+
+class FlaskRestyState(object):
+    def __init__(self, api):
+        super(FlaskRestyState, self).__init__()
```
taion: Remove this – it's just extending `object`

itajaja: :+1:

flask_resty/pagination.py
```diff
@@ -5,9 +5,10 @@
 import sqlalchemy as sa
 from sqlalchemy import Column, sql
 
-from .exceptions import ApiError
 from . import meta
 from . import utils
+from .exceptions import ApiError
```
taion: ?

flask_resty/api.py
itajaja: because `app` could be None there. Also, this works if more than one app is provided during resource  registration

flask_resty/pagination.py
```diff
@@ -63,6 +64,18 @@ def parse_limit(self, limit):
...
         return limit
 
+    def spec_declaration(self, path, spec, **kwargs):
+        path['get'].add_parameter(name='limit',
+                                  type='int',
+                                  description='pagination limit')
+        path['get'].add_property_to_response(
+            prop_name='meta',
+            type='object',
+            properties={
```
taion: This isn't a merge operation, is it? Seems a little weird then – `meta` might contain metadata from multiple other extensions.

flask_resty/pagination.py
itajaja: sorting. Isn't that the correct sorting?

flask_resty/sorting.py
```diff
@@ -37,7 +37,6 @@ def get_criterion(self, view, field_spec):
     def get_column(self, view, field_name):
         return getattr(view.model, field_name)
 
-
```
taion: ?

flask_resty/spec/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Flask-Resty plugin for apispec. Allows passing a Flask-Resty
+`view` to `APISpec.add_path <apispec.APISpec.add_path>`
+"""
+
+from .plugin import setup, schema_path_helper  # noqa
```
taion: Use a top-level

```py
# flake8: noqa
```

flask_resty/spec/declaration.py
```diff
@@ -0,0 +1,84 @@
...
+            for method in ['get', 'post', 'patch', 'delete', 'put']:
+                path[method].add_tag(schema)
+
+
+class ModelViewDeclaration(ApiViewDeclaration):
+    """Declaration for Views that specify a model"""
+
+    def __call__(self, view, path, spec):
+        super(ModelViewDeclaration, self)\
+            .__call__(view, path, spec)
```
taion: You don't need explicit `.__call__` here.

flask_resty/spec/declaration.py
```diff
@@ -0,0 +1,84 @@
...
+
+        for item in [view.pagination, view.filtering, view.sorting]:
```
taion: Use tuple rather than list.

flask_resty/pagination.py
itajaja: You are right here

flask_resty/sorting.py
itajaja: whops

flask_resty/spec/__init__.py
itajaja: :+1:

flask_resty/spec/declaration.py
itajaja: You sure? I tried without, and it didn't seem to work. I'll try again

flask_resty/spec/utils.py
```diff
@@ -0,0 +1,39 @@
+import re
+
+__all__ = ['ref',
```
taion: What's the point of this `__all__`? I think it'd be better to just not use a wildcard import in `__init__`.

flask_resty/spec/declaration.py
itajaja: :+1:

flask_resty/view.py
```diff
@@ -3,12 +3,12 @@
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 from werkzeug.exceptions import NotFound
-
```
taion: ?

tests/test_swagger.py
```diff
@@ -0,0 +1,319 @@
...
+
+from flask_resty import (
+    Api, Filtering, GenericModelView, PagePagination, Sorting,
+    IdCursorPagination,
+)
+from flask_resty.spec import ModelViewDeclaration
+
+# -----------------------------------------------------------------------------
+
+try:
```
taion: Why do you need this if `tox.ini` always installs `apispec`?

flask_resty/spec/utils.py
itajaja: uhm, ok. It really the same thing, after all

flask_resty/spec/declaration.py
taion: I thought so... maybe it doesn't work?

tests/test_swagger.py
itajaja: yes, I don't know where I took this from

taion: I have this for the JWT tests because I run the tests both with and without JWT handling installed.

flask_resty/spec/declaration.py
itajaja: http://stackoverflow.com/questions/12342963/invoking-the-call-method-of-a-superclass

tests/test_swagger.py
itajaja: :+1:

flask_resty/spec/declaration.py
taion: Ah, okay. Thanks.

itajaja:
So I fixed what was to be fixed. Working on changing how a view can overrides the default. I'll let you know when I am done

itajaja:
@taion should be good to go now. Look at last two commits

taion:
OK fix that last comment and then go ahead and merge.

moi90:
This feature should be mentioned in the documentation, shouldn't it? I searched specifically for a way to autogenerate API documentation and found this pull request by accident.

itajaja:
@moi90  flask resty does not support apispec anymore

moi90:
OK, thanks!

PR #79, opened 2016-02-22
[DON'T MERGE] check if item exists before creating missing

itajaja:
Just to start the conversation, would something like that make sense? If so I would add tests and make it better


taion:
This really seems like a concern of the `authorization` to me.

itajaja:
get_item is not using any authorization when creating the item though...

itajaja:
Also, authorization doesn't actually know what's persisted on the database

taion:
I know, but whether it's legal to create a missing item in this context seems like an auth concern.

PR #86, opened 2016-06-08
Allow resources with trailing slashes (DRF style)

### LINKED ISSUE #85: Option to add_resource to use/ not-use trailing slash

cancan101:
Closes #85 
/CC @taion

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -97,6 +97,25 @@ def test_create_client_id(app, views, client):
     }
 
 
+def test_training_slash(app, views, client):
+    api = Api(app)
+    api.add_resource('/widgets/', views['widget_list'], views['widget'])
+
+    response = helpers.request(
+        client,
+        'POST', '/widgets/',
+        {
+            'id': '100',
+        },
+    )
+    assert response.status_code == 201
+    assert response.headers['Location'] == 'http://localhost/widgets/100'
+
+    assert helpers.get_data(response) == {
+        'id': '100',
+    }
+
+
 def test_resource_rules(app, views, client):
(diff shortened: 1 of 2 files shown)


tests/test_misc.py
```diff
@@ -97,6 +97,25 @@ def test_create_client_id(app, views, client):
     }
 
 
+def test_training_slash(app, views, client):
```
taion: Test that the detail URL is correct

taion:
Actually, should we just use `posixpath.join` here?

tests/test_misc.py
```diff
@@ -97,6 +97,27 @@ def test_create_client_id(app, views, client):
...
+            'id': '100',
+        },
+    )
+    assert response.status_code == 201
+    assert response.headers['Location'] == 'http://localhost/widgets/100'
+
+    assert helpers.get_data(response) == {
+        'id': '100',
+    }
+    response = client.get('/widgets/100')
```
taion: Use `id_rule='<id>/'`. Also add a newline before this line.

tests/test_misc.py
```diff
@@ -97,6 +97,28 @@ def test_create_client_id(app, views, client):
...
+            'id': '100',
+        },
+    )
+    assert response.status_code == 201
+    assert response.headers['Location'] == 'http://localhost/widgets/100/'
+
+    assert helpers.get_data(response) == {
+        'id': '100',
+    }
+    response = client.get('/widgets/100/')
```
taion: Add a newline here (since it's a separate test case for the get). Then merge.

PR #87, opened 2016-06-10
Allow `id_fields` to be a Model fields which deals with issues with join when using `filter_by`

cancan101:


taion:
There's plenty of other places where it's expected that `get_id_dict` return something keyed by string...

I'm not sure how you're hitting this and this should include a regression test anyway if possible, but the right way to approach this would be to just do the column name -> column mapping before filtering, since you have access to the model here.

cancan101:
Closing this in favor of something simpler.

PR #88, opened 2016-06-10
Explicit use view model in get_item

cancan101:
Alternative to #87


taion:
This also fails CI...

PR #89, opened 2016-06-14
fix body specification in ApiViewDeclaration

itajaja:


itajaja:
cc @cancan101

cancan101:
:shipit:

PR #90, opened 2016-06-16
Fix create missing semantics

taion:
This changes get_item_or_404 with create_missing to check that the user is authorized to save the item (even if not committing), and raise a 404 if not.

We bypass this check when using the update path with create_missing set, as we will separately authorize saving the created item before committing.

I haven't added full test coverage for both branches of this feature yet. Let me know if this seems reasonable though.

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -183,16 +183,28 @@ class ModelView(ApiView):
 
         return item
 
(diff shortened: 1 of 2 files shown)


taion:
cc @itajaja @cancan101

itajaja:
lgtm

PR #91, opened 2016-06-16
Use SELECT FOR UPDATE

taion:


itajaja:
lgtm

taion:
Closing this out for now – most of our updates are idempotent, so doing SFU by default would not be helpful.

PR #94, opened 2016-07-01
Rename IdCursorPagination to CursorPagination

taion:


taion:
"ID cursor pagination" refers to a different strategy where the cursor only encodes the ID, and the view does an additional lookup to the database to get the sorting-related information from the item.

PR #98, opened 2016-07-01
Add support for pagination meta for items

taion:


taion:
This is what all the earlier PRs lead up to. I need this to fulfill Relay's contract for "smart" create mutations, because I need the cursor on newly created items.

taion:
I've updated the cursor scheme to use shorter cursors – instead of dumping everything to JSON, then base64-encoding, I base64-encode each separate value as a string, remove the padding, then join with the unused `.` character.

In general, this leads to shorter cursors, as I can avoid encoding quotes for strings, and need only a single byte to encode dividers.

This format also opens up the possibility of directly encoding binary values to base64 instead of strings down the road.

flask_resty/view.py
```diff
@@ -266,8 +266,20 @@ def commit(self):
...
 
+    def set_item_meta(self, item):
+        self.set_item_pagination_meta(item)
+
+    def set_item_pagination_meta(self, item):
+        if not self.pagination:
+            return
+
+        pagination_meta = self.pagination.get_item_meta(item, self)
+        if pagination_meta:
```
cancan101: make explcit `None` check

PR #114, opened 2016-10-16
relax assert_value

itajaja:
Either we remove them, or we come up with something more generic. probably we don't need them at all. eg If `actual` is not dictionary-ish enough to have `.get`, it will fail anyway

### DIFF
--- a/flask_resty/compat.py
+++ b/flask_resty/compat.py
@@ -5,5 +5,7 @@ PY2 = int(sys.version_info[0]) == 2
 
 if PY2:
     zip_longest = itertools.izip_longest
+    String = basestring
 else:
     zip_longest = itertools.zip_longest
+    String = str
--- a/flask_resty/compat.py
+++ b/flask_resty/compat.py
@@ -5,7 +5,7 @@ PY2 = int(sys.version_info[0]) == 2
 
 if PY2:
     zip_longest = itertools.izip_longest
-    String = basestring  # noqa
+    basestring = basestring  # noqa
 else:
     zip_longest = itertools.zip_longest
-    String = str
+    basestring = (str, bytes)
--- a/flask_resty/compat.py
+++ b/flask_resty/compat.py
@@ -5,7 +5,7 @@ PY2 = int(sys.version_info[0]) == 2
 
 if PY2:
     zip_longest = itertools.izip_longest
-    String = basestring
+    String = basestring  # noqa
 else:
     zip_longest = itertools.zip_longest
     String = str
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -28,12 +28,13 @@ def get_meta(response):
 
 
 def assert_value(actual, expected):
-    if isinstance(expected, dict):
+    if isinstance(expected, Mapping):
         assert isinstance(actual, Mapping)
         for k, v in expected.items():
             assert_value(actual.get(k, None), v)
-    elif isinstance(expected, Sequence) \
-            and not isinstance(expected, basestring):
+    elif isinstance(expected, basestring):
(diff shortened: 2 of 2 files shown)


itajaja:
...but if we want to solve the failing test, we have to come with a reasonable check :P

itajaja:
maybe checking for `.get` for dicts and `.index` for lists?

taion:
https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping / https://docs.python.org/2/library/collections.html?highlight=mapping#collections.Mapping maybe?

itajaja:
using `Sequence (^ str)` and `Mapping`

flask_resty/testing.py
```diff
@@ -1,3 +1,4 @@
+from collections import Mapping, Sequence
```
taion: this isn't py3 safe is it?

flask_resty/testing.py
```diff
@@ -28,11 +29,11 @@ def get_meta(response):
 
 def assert_value(actual, expected):
     if isinstance(expected, dict):
-        assert isinstance(actual, dict)
+        assert isinstance(actual, Mapping)
         for k, v in expected.items():
             assert_value(actual.get(k, None), v)
-    elif isinstance(expected, list):
-        assert isinstance(actual, list)
+    elif isinstance(expected, Sequence) and not isinstance(expected, str):
```
taion: basestring

itajaja:
g2g?

flask_resty/testing.py
```diff
@@ -28,11 +29,11 @@ def get_meta(response):
 
 def assert_value(actual, expected):
     if isinstance(expected, dict):
-        assert isinstance(actual, dict)
+        assert isinstance(actual, Mapping)
         for k, v in expected.items():
             assert_value(actual.get(k, None), v)
-    elif isinstance(expected, list):
-        assert isinstance(actual, list)
+    elif isinstance(expected, Sequence) and not isinstance(expected, String):
```
taion: this does not match the above where you check for `dict` on `expected`

why do we need to assert on the instance of `actual`, anyway? `[] == {}` is false, right?

flask_resty/compat.py
```diff
@@ -5,5 +5,7 @@
 
 if PY2:
     zip_longest = itertools.izip_longest
+    String = basestring  # noqa
```
itajaja: basestring breaks the 79 chars column in testing! :D

flask_resty/testing.py
taion: No, I mean, why do `elif isinstance(expected, Sequence) and not isinstance(expected, String):` then? Why not just do `isinstance(expected, (list, tuple))` or something? I think you can also just catch the `TypeError` coming from the `zip_longest` as well in the list case.

The dict case looks fine, though.

itajaja: I think my last attempt is a good compromise. The only annoyance is duplicating `assert actual == expected` but I think that's fine.

flask_resty/testing.py
```diff
@@ -28,11 +29,12 @@ def get_meta(response):
 
 def assert_value(actual, expected):
     if isinstance(expected, dict):
-        assert isinstance(actual, dict)
+        assert isinstance(actual, Mapping)
         for k, v in expected.items():
             assert_value(actual.get(k, None), v)
-    elif isinstance(expected, list):
-        assert isinstance(actual, list)
+    elif isinstance(expected, Sequence) \
```
taion: Also, I don't think you need to assert that `actual` is a `Sequence` – the `zip_longest()` will just fail if it isn't, right?

PR #115, opened 2016-11-13
Add exception-catching flush method to ModelView

taion:


flask_resty/view.py
```diff
@@ -258,6 +258,13 @@ def delete_item(self, item):
 
         self.session.delete(item)
 
+    def flush(self):
+        try:
+            # Flushing allows checking invariants without committing.
+            self.session.flush()
+        except IntegrityError:
+            raise ApiError(409, {'code': 'invalid_data.conflict'})
```
itajaja: makes sense to dry it up with commit?

taion: not an easy way to do so – it's small enough that it's not hugely worthwhile to pull that out into a method just to raise an exception imo

PR #118, opened 2016-11-25
Make assert_value stricter

taion:
- Don't support sequences of unequal lengths
- Don't allow ANY to match nonexistent items

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -1,11 +1,11 @@
 from collections import Mapping, Sequence
 import json
 
-from .compat import basestring, zip_longest
+from .compat import basestring
 
 # -----------------------------------------------------------------------------
 
-NO_DATA = object()
+UNDEFINED = object()
 
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


taion:
`None == ANY`, so all of our assertions on `ANY` as dictionary values were no-ops. Likewise with any sequence length checks.

The explicit `UNDEFINED`... well. It's useful!

flask_resty/testing.py
```diff
@@ -30,30 +30,37 @@ def get_meta(response):
 def assert_value(actual, expected):
     if isinstance(expected, Mapping):
         assert isinstance(actual, Mapping)
-        for k, v in expected.items():
-            assert_value(actual.get(k, None), v)
+        # Unlike all the others, this checks that the actual items are a
+        # superset of the expected items, rather than that they match.
+        for key, value in expected.items():
+            if value is not UNDEFINED:
```
itajaja: I don't really think this is the best approach, UNDEFINED shouldn't be a special case handled here, but rather we should have a more stricter ANY that checks for nullness. This code is not aware of any, and it shouldn't be aware of UNDEFINED either

taion: This `UNDEFINED` is a special sentinel for indicating that an item should not be present. There's not really any other way to do this in the code.

There's no way to get `ANY` to do the right thing here, since there's no way to distinguish between a `None` that comes from the caller, and a `None` that comes from a missed `get`.

flask_resty/testing.py
```diff
@@ -30,30 +30,37 @@ def get_meta(response):
...
+                assert_value(actual[key], value)
+            else:
+                assert key not in actual
     elif isinstance(expected, basestring):
         assert actual == expected
     elif isinstance(expected, Sequence):
         assert isinstance(actual, Sequence)
-        for a, e in zip_longest(actual, expected):
-            assert_value(a, e)
+        assert len(actual) == len(expected)
```
itajaja: does this actually change anything? was there something wrong with zip_longest?

flask_resty/testing.py
itajaja: you could have had (lookng at the old code) `assert_value(actual.get(k, UNDEFINED), v)`, so that if v is UNDEFINED, the old code would just behave like this (might be missing something)

PR #119, opened 2016-12-07
Add create_and_add_item hook

taion:


itajaja:
Rationale?

taion:
It much more closely parallels `update_item` and can be overridden as such.

PR #120, opened 2016-12-08
Support resolving IDs to model instances

taion:


taion:
Never mind, this isn't as useful as I thought.

PR #121, opened 2016-12-15
improve testing module

itajaja:
add EVAL testing utilities
rename assert_value to assert_similar

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import re
 
 from dateutil import parser
 
@@ -117,3 +118,14 @@ def _just_before(timestamp):
 
 
 JUST_BEFORE = EVAL(_just_before)
+
+# -----------------------------------------------------------------------------
+
+
+def Matching(expected_regex):
+    regex = re.compile(expected_regex)
+
+    def predicate(actual):
+        return regex.match(actual)
+
+    return EVAL(predicate)
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -1,24 +1,18 @@
-from datetime import datetime, timedelta
 import re
-
-from dateutil import parser
 
 from collections import Mapping, Sequence
 import json
 
 from .compat import basestring
-from .utils import utc
 
 # -----------------------------------------------------------------------------
 
 UNDEFINED = object()
 
-TIME_EPSILON = timedelta(seconds=10)
-
 # -----------------------------------------------------------------------------
 
 
-class EVAL(object):
+class Predicate(object):
     """A helper object to do predicate assertion"""
 
     def __init__(self, predicate):
@@ -51,7 +45,7 @@ def get_meta(response):
     return get_body(response)['meta']
 
 
-def assert_similar(actual, expected):
+def assert_shape(actual, expected):
     if isinstance(expected, Mapping):
         assert isinstance(actual, Mapping)
         # Unlike all the others, this checks that the actual items are a
@@ -59,7 +53,7 @@ def assert_similar(actual, expected):
         for key, value in expected.items():
             if value is not UNDEFINED:
                 assert key in actual
-                assert_similar(actual[key], value)
+                assert_shape(actual[key], value)
             else:
                 assert key not in actual
     elif isinstance(expected, basestring):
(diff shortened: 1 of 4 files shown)


flask_resty/testing.py
```diff
@@ -27,28 +50,39 @@ def get_meta(response):
...
         for key, value in expected.items():
             if value is not UNDEFINED:
                 assert key in actual
-                assert_value(actual[key], value)
+                assert_similar(actual[key], value)
             else:
                 assert key not in actual
     elif isinstance(expected, basestring):
-        assert actual == expected
+        assert expected == actual
```
itajaja: the reason for this is because in python 2, `datetime.now() != JUST_BEFORE`, but `JUST_BEFORE == datetime.now()`, thanks to a crappy python implementation of datetime comparison. fwiw, even `datetime.now() != ANY` which is really bad. python3 works as expected

itajaja: similar thread here: http://osdir.com/ml/python-testing-general/2012-01/msg00025.html

the way they solved it was to invert the order of comparison, like I did here

itajaja:
This is starting to get real fun

flask_resty/testing.py
```diff
@@ -1,12 +1,36 @@
+from datetime import datetime, timedelta
+import re
+
+from dateutil import parser
```
taion: i feel like it's a bit weird to introduce dateutil as a dependency here when it's only used for test support. maybe we can leave out the date assertions?

flask_resty/testing.py
```diff
@@ -1,12 +1,36 @@
...
+
 from collections import Mapping, Sequence
 import json
 
 from .compat import basestring
+from .utils import utc
 
 # -----------------------------------------------------------------------------
 
 UNDEFINED = object()
```
taion: if we had access to `mock` here this could/should be `sentinel.MISSING` or something... but we don't

flask_resty/testing.py
```diff
@@ -1,12 +1,36 @@
...
 
+TIME_EPSILON = timedelta(seconds=10)
+
+# -----------------------------------------------------------------------------
+
+
+class EVAL(object):
```
taion: maybe this should be `Predicate`?

flask_resty/testing.py
```diff
@@ -27,28 +51,39 @@ def get_meta(response):
     return get_body(response)['meta']
 
 
-def assert_value(actual, expected):
+def assert_similar(actual, expected):
```
taion: i thought we were calling this `assert_shape`

flask_resty/testing.py
```diff
@@ -27,28 +51,39 @@ def get_meta(response):
...
-            assert_value(actual_item, expected_item)
+            assert_similar(actual_item, expected_item)
     elif isinstance(expected, float):
         assert abs(actual - expected) < 1e-6
     else:
-        assert actual == expected
+        assert expected == actual
+
+
+def Similar(expected):
```
taion: `Shape`?

flask_resty/utils.py
```diff
@@ -17,3 +22,26 @@ def iter_validation_errors(errors, path=()):
...
+
+# This example is taken from
+# https://docs.python.org/2/library/datetime.html#datetime.tzinfo.fromutc
+
+ZERO = timedelta(0)
+
+HOUR = timedelta(hours=1)
+
+
+class UTC(tzinfo):
```
taion: see comments above re:time stuff

flask_resty/testing.py
itajaja: right I agree. I think all these suite of assertion helpers (Shape, JUST_BEFORE, Matches) could make up a testing library of its own :D

taion: for now, i'd avoid pulling in dateutil here, i think, and look to factor this out either

https://pypi.python.org/pypi/equals looks vaguely similar. `equal`, `equality`, and `eql` are all 404 on PyPI so maybe those would work as names

taion: which ones are it missing?

itajaja: Shape is not exactly like ours, and there is no JUST_NOW. there is Matches, yes

taion: https://github.com/toddsifleet/equals#dictionaries seems very similar to a more explicit version of what we have, no?

itajaja: I don't think it's recursive, to name one difference

taion: right – but that means it's more expressive ,no? though more verbose

taion: explicit*

itajaja: Correct, and that will mean changing all the tests... Meh

taion: Well, what do you want to do? `any_dict.containing(1).not_containing(2)` does feel a bit different from what we have.

Let's stick with this for now, minus the date stuff? And then figure out what to do further later.

itajaja: :+1:

itajaja:
looking good?

taion:
wait, the 100% coverage

taion:
pls

PR #122, opened 2016-12-18
Support returning None from get_location

taion:
This handles the case where created items are not retrievable, for whatever reason.


itajaja:
ah, it happens that I actually needed this as well :)

Are you sure it actually works? doesn't `self.get_location(item)` throws if there is no GET?

taion:
You can override it to return `None`.

itajaja:
ah ok, so this just avoids an empty header. lgtm

taion:
Flask special-cases the `Location` header, so without this you get something like `http://localhost/None` if you try to return `None`.

PR #125, opened 2016-12-23
Add PrimaryKeySorting

taion:
I'm not sure this is a good idea. I was going to use it but then I changed my mind.

### DIFF
--- a/flask_resty/__init__.py
+++ b/flask_resty/__init__.py
@@ -17,7 +17,7 @@ from .pagination import (
     PagePagination, RelayCursorPagination,
 )
 from .related import Related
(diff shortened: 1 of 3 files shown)


itajaja:
Usually FixedSorting would suffice, unless you want to create a base view to apply to different models (with different pk fields). lgtm if you want to add it, just add the spec_declaration

taion:
The nice thing here is that you don't have to type out the primary keys, in case you just want deterministic sorting (e.g. for tests).

taion:
This isn't quite right anyway because it should be by `id_fields`, not by the PK.

PR #127, opened 2016-12-31
Use 'query' instead of 'parameter' in source

taion:
The old name makes no sense.


taion:
Not going to merge this until the next batch of breaking changes – maybe when we cut over to a Blueprint (possibly) for error handling.

taion:
we should add a `query_schema`

taion:
Okay, turns out the old name made lots of sense because it was what the JSON API spec calls for

PR #128, opened 2016-12-31
More flexible filter configuration

taion:


taion:
https://github.com/4Catalyzer/flask-resty/pull/128/commits/ec4a94cdc2dbd3d363f1285856a9b2e0ae7a9637 adds support for customizing empty filter behavior with an API that I think makes sense – though perhaps it's insufficiently explicit? e.g. we could have `empty_filter` and `empty_value` be separate parameters.

I think this resolves the issues we have in our code of handling empty filter values nicely.

taion:
Dunno if we need https://github.com/4Catalyzer/flask-resty/pull/128/commits/34c0bdef153ac91c65ba89291dba89bb2256fa15. Might be too complicated. I can't really see a good use case for it. I'd really only ever use `sql.true()` and `sql.false()`.

flask_resty/filtering.py
```diff
@@ -9,10 +10,18 @@
...
-    def __init__(self, separator=','):
+    def __init__(self, separator=',', empty=''):
         self._separator = separator
+        self._empty = empty
 
     def __call__(self, view, arg_value):
+        if not arg_value:
+            if isinstance(self._empty, ColumnElement):
+                return self._empty
+            elif callable(self._empty):
```
itajaja: this is fine, doesn't hurt :)

flask_resty/filtering.py
```diff
@@ -90,7 +99,13 @@ def make_filter_field(self, arg_name, value):
         elif callable(value):
             return ColumnFilterField(arg_name, value)
         else:
-            return ColumnFilterField(*value)
+            args = tuple(value)
```
itajaja: This is a little bit cryptic, maybe add a comment on what each if handles?

tests/test_filtering.py
```diff
@@ -62,6 +64,13 @@ class WidgetListView(GenericModelView):
 
         filtering = Filtering(
             color=operator.eq,
```
itajaja: I feel like this tuple based api is good only for simple cases, when the filter is complicated, passing first an argument and then a dictionary becomes slightly awkward, it might be more readable to just use `ColumnFilterField` directly? I think that's a less surprising api?

itajaja: then everything could be simplified using kwargs, instead of checking if the first param is a string or a callable

itajaja: eg `Filter(operator.eq)`, `Filter(operator.eq, 'size')`, `Filter(operator.eq, field='size' empty='....')` and so on

taion: `ColumnFilterField` needs to get the column name somehow. If you already build the object, then there's no way for `Filtering` to inject the column name (e.g. what if you share the same `ColumnFilterField` object for two keywords?).

itajaja: One way would be for Filtering to inject it after initialization. You shouldn't share a filter field instance in the same way you shouldn't share a sql alchemy column, the object can only be bound once

taion: nothing prevents you from doing so right now. and i'm pretty sure we do share filter objects, e.g. from decorated functions

taion:
Tried to make the logic clearer.

taion:
I don't really think this is better though.

itajaja:
I think the prev implementation + 2 lines of comment would have been clearer :)

tests/test_filtering.py
itajaja: oki

taion:
Added the comments. Going to merge and make a point release.

PR #129, opened 2017-01-02
use create during update as create

itajaja:
this is just a quickly drafted an idea inside gh editor. I think this more symmetrically matches normal create procedure, especially for authentication. wdyt?

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -108,8 +108,11 @@ def routes(app, models, schemas, auth):
             return self.retrieve(id)
 
     class WidgetCreateMissingView(WidgetViewBase):
-        def create_missing_item(self, id):
-            return self.model(id=id, owner_id=flask.request.args['owner_id'])
+        def create_item(self, data):
+            if 'owner_id' not in data:
+                data['owner_id'] = flask.request.args['owner_id']
+
+            return super(WidgetCreateMissingView, self).create_item(data)
 
         def get(self, id):
             return self.retrieve(id, create_missing=True)
@@ -216,7 +219,7 @@ def test_update_update_missing(client):
             'name': "Created",
         }
     )
-    assert_response(response, 204)
+    assert_response(response, 201)
 
 
 # -----------------------------------------------------------------------------
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -183,7 +183,7 @@ class ModelView(ApiView):
 
         return item
 
-    def get_item(self, id, create_missing=False, for_update=False):
+    def get_item(self, id, create_missing=False):
         try:
             # Can't use self.query.get(), because query might be filtered.
             item = self.query.filter(and_(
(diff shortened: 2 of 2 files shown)


taion:
I don't get it. What if `create_missing` is `False`?

taion:
The idea seems fine though.

itajaja:
Sure, if create missing is false, it should re raise the 404

itajaja:
I'll finalize it tomorrow

itajaja:
I wonder if we could get rid of create_missing_item altogether? is it really critical to support this functionality during get? do we ever use it?

itajaja:
uh, I also have to check for id mismatch I think...

itajaja:
what do you think? I am not 100% sure this is a good idea, but I like that put as create uses the create method, overall. I think it's also more correct to return 201 when inserted in the update method

taion:
Hmm – I don't think 201 makes sense here. `create_missing` is really two different things – one where everything implicitly exists, in which case `create_missing_item` + `update_item` + return 200 is correct; the other where we have upsert semantics, in which case things are different.

Perhaps this should even be a different option. Hmm.

itajaja:
I think 201 is more correct. https://www.ietf.org/rfc/rfc2616.txt states:

> If a new resource is created, the origin server MUST inform the user agent via the 201 (Created) response. If an existing resource is modified, either the 200 (OK) or 204 (No Content) response codes SHOULD be sent to indicate successful completion of the request.

itajaja:
I am not using create_missing anymore though, so the scenario where everything implicitly existis is not the one covered here. it's the one where the resource is actually created

taion:
we have cases where we treat things as implicitly existing, just not (yet) represented in the database because they haven't gotten any non-default values, i think

itajaja:
> we have cases where we treat things as implicitly existing, just not (yet) represented in the database because they haven't gotten any non-default values, i think

Right, but those are not handled in the `update` mehtod. what I am really touching here is the update method, where actually the insert **is** committed. What you are describing is still being handled by `create_missing_item`. I think that, since now `create_missing_item` is only used for the  "resource implicitly exists" case, maybe we should rethink/rename that?

taion:
No, I mean maybe this behavior should have a different name – like `upsert` or something as a kwarg to `update`.

itajaja:
but then that would replace the create_missing kwarg from `update`, otherwise what would be the difference between the two? during update, if the item is created, you **must** return 201. create_missing should just be used for retrieve then

taion:
Because "insert a row into the database" and "create an item" are not the same thing. Suppose we had:

```py
def get(self, id):
    return self.retrieve(id, create_missing=True)
```

I mean maybe the naming isn't ideal, but in some cases we can treat items as _implicitly_ existing even if they don't yet have a db representation.

taion:
What's the status here?

taion:
I think there are a few places where we use `create_missing` when getting an item.

Would it make sense to close this PR?

tests/test_auth.py
```diff
@@ -216,7 +219,7 @@ def test_update_update_missing(client):
             'name': "Created",
         }
     )
-    assert_response(response, 204)
```
itajaja: it's been a while since I looked at this PR, so I might not be able to argument as well. I wonder, tho, why this test expected 204 here. that seems wrong

taion:
The "create missing" path does go through `create_item` now.

PR #135, opened 2017-07-26
Enhance testing

taion:
- Add ApiClient
- Support predicates for assert_response


flask_resty/testing.py
```diff
@@ -97,16 +121,7 @@ def assert_response(response, expected_status_code, expected_data=UNDEFINED):
     else:
         response_data = get_errors(response)
 
-    assert_shape(response_data, expected_data)
-
+    if not isinstance(expected_data, Predicate):
```
itajaja: I am not sure I fully understand the purpose of these two lines. then you should just be able to do `assert response_data = expected_data` then, instead of `assert_shape(response_data, expected_data)`. no?

taion: The idea is that if the user passes in a predicate, then we just use that predicate.

I don't expect this to ever really come up, but it's free.

itajaja: seems like the opposite, this affects all current use cases, since expected_data is NOT usually a predicate, then will be wrapped in a Shape. doing `assert_shape` on a Shape is pointless, isn't it?

taion:
oops

taion:
One last tweak

PR #136, opened 2017-07-26
Stricter linting

taion:


tests/test_auth.py
```diff
@@ -1,13 +1,18 @@
...
-from flask_resty import (
-    Api, ApiError, AuthenticationBase, GenericModelView,
-    HasAnyCredentialsAuthorization, HasCredentialsAuthorizationBase,
-)
-from flask_resty.testing import assert_response
 from marshmallow import fields, Schema
 import pytest
 from sqlalchemy import Column, Integer, String
 
+from flask_resty import (
```
itajaja: don't like this much

taion:
that's why it's marked WIP! i need to tweak the 4c one to not just die when running on py2

tests/test_auth.py
itajaja: the

```py
)
from
```

is particularly ugly compared to `) from`. and also i'd be fine inlining many elements per row. anyhow, I don't really care, was just bitching :P

taion: the nice thing there is, when things break onto multiple lines, if you add something in the middle, you may need to re-wrap the line, which clutters up the diff

itajaja: ture

taion:
Using 4C config now

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+ignore = C813
```
itajaja: why? for py2 compat?

taion: yup; C813 is py3-only trailing commas

itajaja: is it possible to install something like `flake8-config-4catalyzer[py2]` and conditionally trigger rules based on that?

itajaja: I mean, we COULD drop py2 support on FR... :trollface:

taion: I handled it upstream: https://github.com/4Catalyzer/python/pull/16

taion: C813 is stuff in _source_; so things that violate C813 would be illegal syntax in py2

PR #139, opened 2017-10-20
Revamp sorting and cursor pagination

taion:
Use sorting data in pagination if relevant.

### DIFF
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -3,7 +3,6 @@ import base64
 import flask
 from marshmallow import ValidationError
 import sqlalchemy as sa
-from sqlalchemy import Column, sql
 
 from . import meta
 from . import utils
@@ -40,7 +39,7 @@ class LimitPaginationBase(PaginationBase):
         meta.set_response_meta(has_next_page=has_next_page)
         return items
 
-    def get_limit(self, query, view):
+    def get_limit(self):
         raise NotImplementedError()
 
 
@@ -60,9 +59,9 @@ class LimitPagination(LimitPaginationBase):
         self._max_limit = max_limit
 
         if self._max_limit is not None:
-            assert \
-                self._default_limit <= self._max_limit, \
+            assert self._default_limit <= self._max_limit, (
                 "default limit exceeds max limit"
+            )
 
     def get_limit(self):
         limit = flask.request.args.get(self.limit_arg)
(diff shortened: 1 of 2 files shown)


flask_resty/pagination.py
```diff
@@ -184,32 +189,16 @@ class CursorPaginationBase(LimitPagination):
...
+            )
 
         return column_orderings
 
-    def get_column_ordering(self, expression):
-        if isinstance(expression, Column):
-            return expression, True
-
-        column = expression.element
-        assert isinstance(column, Column), "expression is not on a column"
```
itajaja: are these assertions not needed anymore?

flask_resty/pagination.py
```diff
@@ -184,32 +189,16 @@ class CursorPaginationBase(LimitPagination):
     cursor_arg = 'cursor'
 
     def get_column_orderings(self, query, view):
-        column_orderings = tuple(
-            self.get_column_ordering(expression)
-            for expression in query._order_by
-        )
+        column_orderings = view.sorting.get_request_column_orderings(view)
```
itajaja: so you are making it necessary for a view with pagination to have sorting, which is a breaking change. And might not be "strictly" necessary because the ordering might be defined in the model (not saying is a good idea, but it's allowed)

flask_resty/pagination.py
```diff
@@ -184,32 +189,16 @@ class CursorPaginationBase(LimitPagination):
     cursor_arg = 'cursor'
 
     def get_column_orderings(self, query, view):
```
itajaja: should we drop `query`?

flask_resty/pagination.py
```diff
@@ -40,7 +39,7 @@ def get_page(self, query, view):
         meta.set_response_meta(has_next_page=has_next_page)
         return items
 
-    def get_limit(self, query, view):
+    def get_limit(self):
```
itajaja: wait, this was just a typo?

taion:
@itajaja 

The idea here is that, when using a cursor-based pagination, instead of trying to figure out the sorting from the `query`, we just look at `view.sorting`, and ask it how `query` is sorted.

It lets us remove a lot of crummy logic from the cursor pagination implementation, and is more robust to sorting on things like hybrid expressions that aren't columns.

flask_resty/pagination.py
taion: yup

flask_resty/sorting.py
```diff
@@ -7,11 +7,21 @@
 
 class SortingBase(object):
     def sort_query(self, query, view):
+        return self.sort_query_by_fields(
+            query, view, self.get_request_field_orderings(),
```
itajaja: still need `query`?

taion: No, why would I?

itajaja: So let's remove it from the params

taion: wait we still need the query to return the sorted query

itajaja: sorry misread

PR #143, opened 2017-11-20
Add support for multiple keys

jquense:
Overloading the `KEY` arg seems a bit weird but that seems to be not unusual for python? pyjwt does (or will do it) it with audience and issuer so this was a bit more consistent. Maybe we should PR the support up there


tests/keys/testkey_rsa.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDUeDMF8m/Zw6NCvILB7w0R6WuI5M0VYplSK969StGGWpyZkl0K+4IlGv7EI346Y7yrCAklH3x9426PFWD4n0uzf9Q4DgdwX2dgfpqpVcQsATweT8Au/GL3qazUMO4C9AZphn6UA3GW/F5AU5lJuOTP/2aHxot4YrCokunen46FQFCnC8BQA+Dzb45S2BNbedKWON4WvXgHyoCd4ZJz/CXSeL77VJFlRcVKi/NypA6SVrdbbuust+sPoteCP9OwOzi8PU+cibXNxV6cpPtN9gjXtuwb879Z5zDPY2XX2kvuAOiw2u4XeiAlY2rUvKf+KM9CeQ8ulkfKG+6LJxhwjcOJ aasmundo@mair.local
```
jquense: taken from the pyjwt test suite

flask_resty/jwt.py
```diff
@@ -42,7 +42,7 @@ def get_request_credentials(self):
 
         try:
             payload = self.decode_token(token)
-        except InvalidTokenError:
+        except InvalidTokenError as e:
```
itajaja: revert

flask_resty/jwt.py
```diff
@@ -70,8 +70,25 @@ def get_token_from_authorization(self, authorization):
     def get_token_from_request(self):
         return flask.request.args.get(self.id_token_arg)
 
+    def get_key_for_token(self, keys, token):
+        unverified_header = jwt.get_unverified_header(token)
+
+        try:
+            token_kid = unverified_header['kid']
+        except KeyError:
+            raise InvalidTokenError('Key ID header parameter is missing')
```
itajaja: I would do `raise ApiError(401, {'code': 'invalid_token.missing_kid'})` or something like that otherwise it would be hard to debug

flask_resty/jwt.py
```diff
@@ -70,8 +70,25 @@ def get_token_from_authorization(self, authorization):
...
+
+        for kid, key in keys.items():
+            if kid == token_kid:
+                return key
+
```
itajaja: raise here

setup.py
```diff
@@ -55,6 +55,7 @@ def run(self):
     ),
     extras_require={
         'jwt': ('PyJWT >= 1.4.0',),
+        'cryptography': ('cryptography >= 2.0.3',),
```
itajaja: this is only used in tests right? why putting it here

flask_resty/jwt.py
jquense: i wasn't sure if the lack of specific error generally around `decode` was intentional to not leak details?

setup.py
jquense: only tests. it's here because i had no idea where to put it

taion: if it's test-only, then you don't need it here at all; tox takes care of it

tox.ini
```diff
@@ -12,7 +12,7 @@ deps =
     pytest-cov
 
 commands =
-    jwt: pip install -e .[jwt]
+    jwt: pip install -e .[cryptography,jwt]
```
taion: add `jwt: cryptography` to the `deps` section above per http://tox.readthedocs.io/en/latest/config.html#generating-environments-conditional-settings

flask_resty/jwt.py
taion: imo instead of doing this, we should have a separate `JwksAuthentication` that gets configured with the JWKS

taion:
`JwkSetAuthentication` and maybe `JwkAuthentication`

taion:
(JWKS is a terrible acronym that AFAICT is never used in the RFC)

setup.py
jquense: it wasn't as far as I could see. I need to add the install step to the tox.ini?

tox.ini
jquense: niiice ok

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
...
+    def __init__(self, key_set=None, **kwargs):
+        super(JwkSetAuthentication, self).__init__(**kwargs)
+
+        self.jwks = key_set
+        self.algorithms = get_default_algorithms()
+
+    def get_jwks(self):
+        config = flask.current_app.config
+        return self.jwks \
+            if self.jwks else config[self.get_config_key('key_set')]
```
chirag04: need to add key_set to `JWT_DECODE_ARG_KEYS`?

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
 
     def get_credentials(self, payload):
         return payload
+
+
+class JwkSetAuthentication(JwtAuthentication):
+    def __init__(self, key_set=None, **kwargs):
+        super(JwkSetAuthentication, self).__init__(**kwargs)
+
+        self.jwks = key_set
```
taion: this should be `self.jwk_set`

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
 
     def get_credentials(self, payload):
         return payload
+
+
+class JwkSetAuthentication(JwtAuthentication):
+    def __init__(self, key_set=None, **kwargs):
```
taion: maybe `jwk_set` instead of `key_set`?

flask_resty/jwt.py
taion: no i think this is ok

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
...
+            ).public_key()
+
+        # awkward
+        return algorithm.from_jwk(json.dumps(jwk))
+
+    def get_key_for_token(self, token):
+        unverified_header = jwt.get_unverified_header(token)
+
+        try:
+            algorithm = self.algorithms[unverified_header['alg']]
```
taion: do not use `alg` from here – we can't trust it. instead use `alg` on the JWK. this is just going to have to be a little bit different. we trust `alg` on the JWK but not on the JWT

(though perhaps we should fail if there's a mismatch between the two `alg` fields)

flask_resty/jwt.py
taion: i'm not sure how this comment got carried over, but `invalid_token.missing_kid` sounds good

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
...
+            raise InvalidTokenError('Algorithm header parameter is missing')
+
+        try:
+            token_kid = unverified_header['kid']
+        except KeyError:
+            raise InvalidTokenError('Key ID header parameter is missing')
+
+        for jwk in self.get_jwks()['keys']:
+            if jwk['kid'] == token_kid:
+                return self.from_jwk(jwk, algorithm)
```
taion: again `algorithm` should come from the JWK, not the JWT

flask_resty/jwt.py
taion: oh or does this get handled upstream?

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
...
+
+        raise InvalidTokenError("no key found")
+
+    def decode_token(self, token):
+        args = self.get_jwt_decode_args()
+
+        args['key'] = self.get_key_for_token(token)
+        print(jwt.decode(token, **args))
```
taion: remove

flask_resty/jwt.py
```diff
@@ -89,3 +96,53 @@ def get_config_key(self, key):
...
+
+        self.jwks = key_set
+        self.algorithms = get_default_algorithms()
+
+    def get_jwks(self):
+        config = flask.current_app.config
+        return self.jwks \
+            if self.jwks else config[self.get_config_key('key_set')]
+
+    def from_jwk(self, jwk, algorithm):
```
taion: maybe call this `key_from_jwk`?

tox.ini
taion: fix?

tox.ini
```diff
@@ -12,7 +12,7 @@ deps =
...
 
     flake8 .
-    pytest --cov
+    pytest {posargs:--cov}
```
taion: eh?

setup.py
taion: remove

jquense: Needed now, for cert loading

flask_resty/jwt.py
jquense: Not sure what to do here tho as `alg` is optional on the jwk (pyjwt doesn't even add it). I think in the upstream PR I raise on a mismatch if both are present? That may only be in the encode step tho. Also this what pyjwt does :/

setup.py
taion: oh, then make it part of `jwt`

flask_resty/jwt.py
taion: it's really unsafe to read it from the header, though. the user controls this value, and it could easily be "no signing" or something else totally screwed up

jquense: there is a safeguard, the token value is checked against the decode args algorithm, if the user changes it on the token verification will fail, so you cant change it to 'none' unless your server supports that, which would be dumb

yodaiken: did you see the auth0 none vuln from a few years ago? what do you think about just requiring an alg in the JWK, or defaulting to a user customizable value (default RS256)?

jquense: i think on our end we can just require `alg` be in the jwk, but as a generic jwk authentication, that may be too opinionated? IDK, we can certainly require that the config at leasts provides the algo whitelist, that should avoid this vulnerability right? PyJWT sort of requires it, it warns that not providing it is deprecated.

tox.ini
jquense: then i run `tox -- -vx -k jwt_set` and pass through the flake args

flask_resty/jwt.py
```diff
@@ -89,3 +96,59 @@ def get_config_key(self, key):
 
     def get_credentials(self, payload):
         return payload
+
+
+class JwkSetAuthentication(JwtAuthentication):
+    def __init__(self, jwk_set=None, **kwargs):
+        super(JwkSetAuthentication, self).__init__(**kwargs)
+
+        self.jwks = jwk_set
```
taion: `self.jwk_set`

flask_resty/jwt.py
```diff
@@ -89,3 +96,59 @@ def get_config_key(self, key):
...
+        self.algorithms = get_default_algorithms()
+
+    def get_jwks(self):
+        config = flask.current_app.config
+        return self.jwks \
```
taion: (formatting nit) write as

```python
return (
    self.jwk_set if self.jwk_set
    else config[self.get_config_key('jwk_set')]
)
```

flask_resty/jwt.py
```diff
@@ -89,3 +96,59 @@ def get_config_key(self, key):
...
+
+        raise InvalidTokenError("no key found")
+
+    def decode_token(self, token):
+        args = self.get_jwt_decode_args()
+
+        unverified_header = jwt.get_unverified_header(token)
+        jwk = self.get_jwk_for_token(token)
+
+        alg = jwk['alg'] if 'alg' in jwk else unverified_header['alg']
```
taion: Ah, I see. Can you add a comment here to the effect of:

> It's safe to use the alg from the header here, as we verify that against the algorithm whitelist later.

flask_resty/jwt.py
```diff
@@ -89,3 +96,59 @@ def get_config_key(self, key):
...
+
+        # I'm not sure we need to re-up this check here, jwt.decoode will
```
taion: "jwt.decode"

taion: i think this is a good check

setup.py
```diff
@@ -54,7 +54,7 @@ def run(self):
         'Werkzeug >= 0.11',
     ),
     extras_require={
-        'jwt': ('PyJWT >= 1.4.0',),
+        'jwt': ('PyJWT >= 1.4.0', 'cryptography >= 2.0.3'),
```
taion: Maybe `>= 2.0.0`? All Python deps are peer deps, so it's nice to use round-ish numbers.

tests/test_jwk_set.py
```diff
@@ -0,0 +1,205 @@
+import json
```
taion: You could merge this into `test_jwt`. Just add another view in the same fixture set.

flask_resty/jwt.py
```diff
@@ -89,3 +96,59 @@ def get_config_key(self, key):
...
+        # check and raise the the same thing
+        if alg not in args['algorithms']:
+            raise InvalidAlgorithmError(
```
taion: Would you mind getting a coverage hit on here? Just try a token with `"alg": "none"`. It's a good check to have.

f.py
```diff
@@ -0,0 +1,22 @@
+from base64 import b64decode
```
taion: remove this file?

ff.py
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
```
taion: remove this file?

tox.ini
```diff
@@ -14,5 +14,5 @@ deps =
 commands =
     jwt: pip install -e .[jwt]
 
-    flake8 .
-    pytest --cov
+    # flake8 .
```
taion: fix

f.py
jquense: arggg how did get in

jquense:
release?

tox.ini
```diff
@@ -15,4 +15,4 @@ commands =
     jwt: pip install -e .[jwt]
 
     flake8 .
-    pytest --cov
+    pytest {posargs:--cov}
```
taion: btw why not

```
--cov {posargs}
```

or something similar where we don't disable `--cov` when using positional args?

jquense: well when trying it locally i don;t usually want coverage to run, if i'm just trying to run tests on a quick cycle

taion: Does it meaningfully slow things down? You can also explicitly disable coverage on CLI

PR #144, opened 2017-11-22
Tweak the JWK set code a bit

taion:


flask_resty/jwt.py
```diff
@@ -142,16 +142,18 @@ def decode_token(self, token):
...
+                "The specified alg value is not allowed",
             )
 
-        args['key'] = self.key_from_jwk(jwk, self.algorithms[alg])
-
-        return jwt.decode(token, **args)
+        return jwt.decode(
+            token,
+            key=self.get_key_from_jwk(jwk, self.algorithms[alg]),
+            **args
```
jquense: does this get overridden if you provided a key in the config somewhere?

taion:
```
In [1]: def foo(**kwargs): pass

In [2]: foo(key=1, **{'key': 2})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-74b1f61e9cb5> in <module>()
----> 1 foo(key=1, **{'key': 2})

TypeError: foo() got multiple values for keyword argument 'key'
```

And this will correspondingly be a 500, which I think is what we want.

jquense: nice. 👍

PR #145, opened 2017-11-28
Allow multiple issuers and audiences

jquense:
Already upstream PR’s for this

### DIFF
--- a/flask_resty/jwt.py
+++ b/flask_resty/jwt.py
@@ -107,10 +107,10 @@ class JwtAuthentication(AuthenticationBase):
 
 class JwkSetAuthentication(JwtAuthentication):
     def __init__(self, jwk_set=None, options=None, **kwargs):
-        if options:
-            options.update({'verify_aud': False, 'verify_iss': False})
-        else:
-            options = {'verify_aud': False, 'verify_iss': False}
+        if not options:
+            options = {}
+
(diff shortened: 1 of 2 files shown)


flask_resty/jwt.py
```diff
@@ -49,7 +55,8 @@ def get_request_credentials(self):
 
         try:
             payload = self.decode_token(token)
-        except InvalidTokenError:
+        except InvalidTokenError as e:
+            print(e)
```
taion: remove

jquense:
I guess logically yeah, I stuck it here because it come up with sets, of different keys, tokens

taion:
I'd move it upstream and merge. That or how responsive are the PyJWT devs on accepting and releasing external changes?

jquense:
I haven't heard anything from the pyJWT folks in a week on any of my PR's so IDK. by upstream here you mean...to the JWT authentication? We can also just hold off for now and see if pyJWT merges it, its not hard to handle further down on my end until then

taion:
That's a little annoying that they're not responding on such a small change.

If this is a blocker for you, I'd say let's move this into the parent JWT auth, and we can merge and just cut a new (non-breaking) release that uses the updated PyJWT when that's possible.

taion:
@jquense Was this upstreamed into PyJWT?

jquense:
no real feedback from them they on a glacial pace there

PR #146, opened 2017-11-28
Make status_code on add_ping configurable

taion:


taion:
@itajaja Done. Take a look?

PR #147, opened 2017-12-20
Roll back database changes after problems

taion:


taion:
Maybe the rollback here isn't ideal. We could test `session.is_active` in the test client. It seems a little weird to handle the rollback for the user.

Expunging items that received unauthorized changes seems like a free win, though.

taion:
Not going ahead with this.

PR #148, opened 2017-12-20
Expunge instances with unauthorized changes

taion:
Just in case someone catches the exception and commits, we don't want to accidentally commit bad changes.

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -268,7 +268,11 @@ class ModelView(ApiView):
     def add_item(self, item):
         self.session.add(item)
 
-        self.authorization.authorize_save_item(item)
+        try:
+            self.authorization.authorize_save_item(item)
+        except Exception:
+            self.session.expunge(item)
+            raise
 
     def create_and_add_item(self, data):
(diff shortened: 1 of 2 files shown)


taion:
Expunge is quick – it's just detaching the instance from the session.

The concern is like, if you catch the API error to rethrow it, and commit or something in the meantime, you'd end up committing the dirty instances that failed authZ, which seems like a bad idea.

This isn't actually something we ever do, but it seems like a nice guardrail.

itajaja:
just defensive then? why not catching the whole `update_item` and `add_item` then?

itajaja:
mm I guess my comment doesn't make sense -- I like the view code being really terse, that's to me is on of the major winning point of flask-resty. Besides that, I am fine with this change

taion:
I dunno, I guess this isn't super duper urgent. You think this isn't worth adding?

itajaja:
up to you, I don't think it's super worth, hopefully if we accidentally commit we have unit tests for that :)

taion:
That's a fair point. I won't add the expunge for now then.

PR #149, opened 2017-12-20
Format HTTP errors as JSON

taion:
Apparently I just had this sitting around. Not sure why I didn't PR this.


itajaja:
can we cover it?

taion:
Oh, it's because this code didn't work. Fixed and tested.

taion:
Okay this will be breaking.

PR #152, opened 2017-12-28
Support args schemas

taion:


taion:
This is a pretty trivial change. It's just to allow a slightly easier way of deserializing query parameters.

tests/test_query.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_error_get_name_missing(client):
+    response = client.get('/name')
+    assert_response(response, 422, [{
+        'code': 'invalid_query',
+        'source': {'parameter': 'name'},
+    }])
+
+
+def test_error_get_name_many(client):
```
taion: This is not the default behavior from Flask when reading query args, but I think it's more correct to fail on requests that specify a query parameter more than once when not expected.

flask_resty/view.py
```diff
@@ -18,6 +18,7 @@
 class ApiView(MethodView):
     schema = None
     id_fields = ('id',)
+    query_schema = None
```
itajaja: can we call it `querystring_schema` or `query_string_schema` to avoid semantic collision with sql query?

tests/test_query.py
```diff
@@ -0,0 +1,90 @@
...
+    response = client.get('/name?name=foo&name=bar')
+    assert_response(response, 422, [{
+        'code': 'invalid_query',
+        'source': {'parameter': 'name'},
+    }])
+
+
+def test_error_get_names_missing(client):
```
itajaja: this is a little bit odd, since the semantic of `required` on a list is never that the list has to be non empty. I think it would make more sense if `required` in this context didn't have any effect. if you want the non empty semantic you can use a custom validator

flask_resty/view.py
```diff
@@ -144,6 +145,47 @@ def get_data_id(self, data):
...
+        for field_name, field in self.query_schema.fields.items():
+            if field_name in args:
+                args_key = field_name
+            elif field.load_from and field.load_from in args:
+                args_key = field.load_from
+            else:
+                continue
+
+            value = args.getlist(args_key)
+            if not self.is_list_field(field) and len(value) == 1:
```
itajaja: I would probably inline it instead of having an `is_list_field` method on the view

flask_resty/view.py
itajaja: In general, all the query things that talk about querystring. not a big deal

flask_resty/view.py
taion: it's to allow overriding, just in case

tests/test_query.py
taion: what do you mean? the parameter is missing entirely, right?

flask_resty/view.py
taion: maybe `args_schema`? not the clearest name, but consistent with what flask calls it otherwise. `query_string` looks clunky

tests/test_query.py
taion: the user could always set `missing=[]` or something

flask_resty/view.py
taion: `qs_schema`?

taion: by analogy with https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs

flask_resty/view.py
itajaja: why do you think anyone would need to override this tho?

flask_resty/view.py
itajaja: json api calls it `parameters` :D

tests/test_query.py
itajaja: uh, isn't the way of defining lists `foo=1&foo=2` and not `foo=[...]`? in the former case, there is no distinction between "empty" and "missing".

flask_resty/view.py
taion: “parameter”, for a single parameter, in the error, no?

flask_resty/view.py
taion: like if they have a different list field

tests/test_query.py
taion: well, querystring does, no? you could always do `?foo`

flask_resty/view.py
itajaja: yeah, I mean, nobody knows how to call them... `args` or `qs` sound good

flask_resty/view.py
itajaja: wouldn't that be a subclass of list most probably? if you feel that can happen anyway, 👍

tests/test_query.py
itajaja: that's also ambiguous, because it could be interpreted as `[None]`.

flask_resty/view.py
taion: or if it's not marshmallow... you can actually override all the marshmallow hooks here i think

tests/test_query.py
taion: well, it's not ideal, but i think i'm ok with matching existing semantics... if you call `args.getlist()` on something missing, you do get an error

taion:
Updated

tests/test_query.py
itajaja: I guess there isn't an optimal solution. ok with it

itajaja:
also, what about more complicated objects like dictionaries/maps etc. should we leave it as unspecified?

taion:
Unspecified and not supported. Just like query-string. 

I’ll add a test case for `missing`.

PR #153, opened 2018-01-20
Add hook for authorizing creates

taion:


flask_resty/authorization.py
```diff
@@ -14,6 +14,9 @@ def authorize_request(self):
     def filter_query(self, query, view):
         raise NotImplementedError()
 
+    def authorize_create_item(self, item):
+        raise NotImplementedError()
```
taion: Should this default to `pass`? That would make this change non-breaking, but it feels sort of weird.

itajaja: good question. I vote for the breaking change, all our application will catch this in the unit test anyway

taion: Yeah, that's my inclination right now.

flask_resty/view.py
```diff
@@ -308,6 +308,8 @@ def create_item(self, data):
         return self.model(**data)
 
     def add_item(self, item):
+        self.authorization.authorize_create_item(item)
```
taion: Should this be called `authorize_add_item` instead? We're not actually calling it in `create_item`, but instead in `add_item`. But `authorize_create_item` sounds better.

itajaja: create is more meaningful than add

PR #155, opened 2018-01-20
Make SortingBase just the sorting contract

taion:


taion:
Technically, yeah. In practice no :p

PR #156, opened 2018-01-21
Revamp filtering

taion:


taion:
The old API was an abominable mess. This is a bit cleaner.

tests/test_filtering.py
```diff
@@ -244,3 +294,46 @@ def test_error_invalid_field(client):
...
+        'code': 'invalid_filter.missing',
+        'source': {'parameter': 'size'},
+    }])
+
+
+def test_error_model_filter_required_missing(client):
+    response = client.get('/widgets_color_custom')
+    assert_response(response, 400, [{
+        'code': 'invalid_filter',
+        'detail': 'Missing data for required field.',
```
taion: This is inconsistent with the above. A bit annoying. I think I want to move the parts of the code after the dots to an `extra` or a `meta.extra` field.

tests/test_filtering.py
```diff
@@ -42,46 +48,70 @@ class WidgetSchema(Schema):
...
 def routes(app, models, schemas, filter_fields):
-    class WidgetListView(GenericModelView):
+    class WidgetViewBase(GenericModelView):
         model = models['widget']
         schema = schemas['widget']
 
+    class WidgetListView(WidgetViewBase):
         filtering = Filtering(
             color=operator.eq,
-            size=(operator.eq, {
```
taion: This syntax was stupid I don't know why I thought this was a good idea.

flask_resty/filtering.py
```diff
@@ -52,79 +79,119 @@ def format_validation_error(self, message):
...
-    def get_filter_clause(self, view, value):
-        return self._filter(view.model, value)
+    def get_filter_arg(self, view):
+        return view.model
+
+
+# -----------------------------------------------------------------------------
+
+
+def model_filter(field, **kwargs):
```
taion: This used to be `@filter_function`. I'm 60/40 on renaming this. This is the only part of the API that's changing that we actually use. Nothing else here imposes changes on most application code.

taion:
Still needs a bit of work. The handling for `empty` is questionable. It should probably just control either "pass through" or "return no rows".

taion:
Good to go now. Instead of having a very generic `empty` parameter that does 3 different things, I just have a single `allow_empty` parameter now. The behavior we use in many places of returning empty lists for empty filter values is now the default (so no more `empty=sql.false()`), but we can explicitly set `allow_empty=True` for cases where empty is a legitimate filter value.

jquense:
not sure i understand ll of this, but LGTM. The functinal API change is that there is now a proper set of classes to define filter behavior, e.g. ColumnFilter?

taion:
One more:

- Add `validate` parameter to `ColumnFilter` that disables field validation when set to `False`

jquense:
:shipit:

taion:
@itajaja Thoughts on the new `validate` flag for `ColumnFilter`?

PR #157, opened 2018-01-21
Add FOR UPDATE support

taion:


jquense:
i'm still not sure i understand what `with_for_update` does...

itajaja:
from [doc](https://www.postgresql.org/docs/9.0/static/sql-select.html#SQL-FOR-UPDATE-SHARE)

> FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being modified or deleted by other transactions until the current transaction ends.

we use it in qsi in at least one place

PR #160, opened 2018-01-24
Add AuthorizeModifyMixin

taion:


taion:
Need to add a test that the actions look right.

taion:
Yeah, it's just a shorthand. We use this pattern pretty much everywhere, though, even in non-tenant cases where the Flask-RESTy-Tenants stuff isn't an appropriate base class.

itajaja:
lgtm after you  add tests

PR #165, opened 2018-01-31
feat: Allow disabling appending slash

taion:
ref https://github.com/pallets/werkzeug/issues/1246


taion:
Starting to think this is the wrong approach. Makes more sense to just use `StrictRule` as `rule_class` in app. There's no benefit in putting this on `Api`.

taion:
d'oh, wait, i didn't read my own darn comment

PR #168, opened 2018-02-06
Handle uncaught app errors

taion:


flask_resty/api.py
```diff
@@ -18,6 +22,11 @@ def handle_api_error(error):
 
 
 def handle_http_exception(error):
+    # Flask calls the InternalServerError handler with any uncaught app
+    # exceptions. Re-raise those as generic internal server errors.
+    if not isinstance(error, HTTPException):
+        error = InternalServerError()
```
itajaja: I wonder, where/when is the original error's stacktrace printed? This modification wouldn't change that right? In the sense that the original error's stacktrace should still be printed

taion: Apparently it still gets printed. I have no idea why.

taion: It's here: https://github.com/pallets/flask/blob/f9c6f389ac71b1354a07925785ea4738bf5ba496/flask/app.py#L1701

PR #169, opened 2018-02-21
improve sorting

itajaja:
I'll add tests when this looks good. This is not super flexible, as the fixed fields are only added in the end. The alternative would be to change the api to looks something like `Sorting('a', '-a', '+b')` where the one with +/- are fixed. bleah. other ideas?

Alternatively we can just move the ApiError check to `get_request_field_orderings` like in this PR, and this would help with easier subclassing

### DIFF
--- a/flask_resty/sorting.py
+++ b/flask_resty/sorting.py
@@ -63,25 +63,27 @@ class FixedSorting(FieldSortingBase):
 class Sorting(FieldSortingBase):
     sort_arg = 'sort'
 
-    def __init__(self, *field_names, **kwargs):
+    def __init__(self, *field_names, fixed_fields=None, **kwargs):
         self._field_names = frozenset(field_names)
+        self._fixed_orderings = ()
+        if fixed_fields:
+            self._fixed_orderings = self.get_field_orderings(fixed_fields)
         self._default_sort = kwargs.get('default')
 
     def get_request_field_orderings(self):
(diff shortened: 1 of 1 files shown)


itajaja:
see alternative at https://github.com/4Catalyzer/flask-resty/pull/170

taion:
I don't understand why #170 is an alternative. Isn't it just a subset of this PR?

taion:
I wonder if we could use something like:

```python
class IdSortingContainer(FieldSortingBase):
    def __init__(self, sorting):
        self._sorting = sorting

        query = self._sorting.sort_query(query, view)

        id_field_orderings = ((id_field, True) for id_field in view.id_fields)
        return self.sort_query_by_fields(query, view, id_field_orderings)

    def get_request_field_orderings(self):
        # TODO: Explicit tuple casts?
        return self._sorting.get_request_field_orderings() + foo  # ?? view
```

Not sure how exactly I want this API to look, but I think extending an existing sorting object to sort by ID fields is more like what we want.

itajaja:
> I don't understand why #170 is an alternative. Isn't it just a subset of this PR?

That's correct. I just wanted to distinguish the two. I am not fond of this tho. I would propose to not add additional sorting classes in FR, and just allow more flexibility in userland

itajaja:
kthxbye

PR #171, opened 2018-02-22
Automatically sort by ID in cursor pagination

taion:
cc @itajaja

I think this is a more direct way at #169.

### DIFF
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -188,21 +188,52 @@ class PagePagination(LimitOffsetPagination):
 class CursorPaginationBase(LimitPagination):
     cursor_arg = 'cursor'
 
+    def ensure_query_sorting(self, query, view):
+        sorting_field_orderings, missing_field_orderings = (
+            self.get_sorting_and_missing_field_orderings(view)
+        )
+
+        query = view.sorting.sort_query_by_fields(
+            query,
+            view,
+            missing_field_orderings,
+        )
+        field_orderings = sorting_field_orderings + missing_field_orderings
+
+        return query, field_orderings
+
     def get_field_orderings(self, view):
+        sorting_field_orderings, missing_field_orderings = (
+            self.get_sorting_and_missing_field_orderings(view)
+        )
+        return sorting_field_orderings + missing_field_orderings
+
+    def get_sorting_and_missing_field_orderings(self, view):
         sorting = view.sorting
         assert sorting is not None, (
             "sorting must be defined when using cursor pagination"
         )
 
-        field_orderings = sorting.get_request_field_orderings()
-        for id_field in view.id_fields:
-            assert (
-                id_field in (field_name for field_name, _ in field_orderings)
-            ), (
-                "ordering does not include {}".format(id_field)
-            )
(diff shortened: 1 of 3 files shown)


flask_resty/sorting.py
```diff
@@ -15,14 +15,14 @@ def sort_query(self, query, view):
...
-        field_orderings = self.get_request_field_orderings()
+        field_orderings = self.get_request_field_orderings(view)
         return self.sort_query_by_fields(query, view, field_orderings)
 
     def sort_query_by_fields(self, query, view, field_orderings):
         criteria = self.get_criteria(view, field_orderings)
         return query.order_by(*criteria)
 
-    def get_request_field_orderings(self):
+    def get_request_field_orderings(self, view):
```
taion: This change is unnecessary, but I think it's probably better to have this anyway.

tests/test_pagination.py
```diff
@@ -396,45 +396,89 @@ def test_relay_cursor_default(client):
...
         'cursors': [
             'MQ.NA',
-            'MQ.MQ',
+            'Mg.Mg',
+        ],
+    }
+
+
+def test_relay_cursor_sorted_inverse(client):
+    response = client.get('/relay_cursor_widgets?sort=-size&cursor=Mg.NQ')
```
taion: Note that this gives `-size,-id`, not `-size,id`.

flask_resty/pagination.py
```diff
@@ -188,21 +188,52 @@ def spec_declaration(self, path, spec, **kwargs):
 class CursorPaginationBase(LimitPagination):
     cursor_arg = 'cursor'
 
+    def ensure_query_sorting(self, query, view):
```
itajaja: I think it's cleaner if instead of returning a tuple `get_page` first calls `field_orderings  = get_field_orderings` and then `query = ensure_query_sorting()`

flask_resty/pagination.py
```diff
@@ -188,21 +188,52 @@ def spec_declaration(self, path, spec, **kwargs):
...
-                "ordering does not include {}".format(id_field)
-            )
+        sorting_field_orderings = sorting.get_request_field_orderings(view)
 
-        return field_orderings
+        sorting_ordering_fields = frozenset(
+            field_name for field_name, _ in sorting_field_orderings,
+        )
+
+        # For convenience, use the ascending setting on the last explicit
```
itajaja: why is this better than just setting it to `True` every time?

flask_resty/pagination.py
taion: it would be, but i couldn’t figure out a good way to tell which orderings  needed to be added in that case. we need to update both the query and the orderings, and that logic isn’t specific to the relay scheme

flask_resty/pagination.py
taion: if you do `-size`, it’d be nice to get the reverse

flask_resty/sorting.py
taion: @itajaja Thoughts on whether this change makes sense?

itajaja: I guess it's ok, also ok to add it later when needed. up to you

taion:
v0.15.0

PR #173, opened 2018-03-19
Support related IDs

taion:
As discussed.


taion:
Thoughts?

taion:
@jquense They're necessary to enable this, because in this case the "in" key and the "out" key are different.

PR #176, opened 2018-04-09
Fix compatibility with newer apispec

taion:


flask_resty/spec/plugin.py
```diff
@@ -30,7 +30,7 @@ def schema_path_helper(spec, path, view, **kwargs):
             'type': 'string',
         })
     if parameters:
-        path['parameters'] = parameters
+        operations['parameters'] = parameters
```
taion: This ends up at the same place... but why? Why is `parameters` a key under operations?

PR #178, opened 2018-04-12
Will this work?

jquense:


flask_resty/view.py
```diff
@@ -348,12 +348,14 @@ def flush(self):
             # Flushing allows checking invariants without committing.
             self.session.flush()
         except IntegrityError:
+            flask.current_app.logger.exception("Integrity error")
```
taion: the stack trace will say `IntegrityError`, so can you make the message here like `"flush failed"`?

flask_resty/view.py
```diff
@@ -348,12 +348,14 @@ def flush(self):
...
             raise ApiError(409, {'code': 'invalid_data.conflict'})
 
     def commit(self):
         try:
             self.session.commit()
         except IntegrityError:
+            flask.current_app.logger.exception("Integrity error")
```
taion: `"commit failed"`?

flask_resty/view.py
```diff
@@ -348,12 +348,14 @@ def flush(self):
             # Flushing allows checking invariants without committing.
             self.session.flush()
         except IntegrityError:
+            flask.current_app.logger.exception("flush failed")
```
taion: Thinking about this a little more, I'm actually a bit scared of this. This opts users into logging potentially sensitive data to logs that may otherwise not have them (since errors are usually logged).

It also doesn't quite give us what we want, either – we want to see these end up in Sentry, not in the app logs.

PR #180, opened 2018-04-12
Use blinker to signal handled integrity errors

taion:
I dunno. What do you think? This feels more semantically correct.

Sentry also lets us capture log events, but... I'm not sure. It seems better to capture an exception than a log message anyway, no?

But actually Sentry has some clever stuff to handle logging, and using a logger is more consistent with what we're doing on the JS side.

But here, the idiomatic way to scope loggers is by module, and that seems insufficiently semantically meaningful.

Thoughts?

### DIFF
--- a/flask_resty/__init__.py
+++ b/flask_resty/__init__.py
@@ -29,6 +29,7 @@ from .pagination import (
     RelayCursorPagination,
 )
 from .related import Related, RelatedId
+from .signals import got_handled_integrity_error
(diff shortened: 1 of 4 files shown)


jquense:
I'm not best judge of waht is more correct in the python/flask context. I do want to have the option to see integrity errors in stderr like we see exceptions tho, for our k8s logging

taion:
eh, seems like we don't want this after all

PR #183, opened 2018-05-14
Add explicit support for query options

taion:


flask_resty/view.py
```diff
@@ -209,11 +212,47 @@ def session(self):
...
+    #: Base query options to apply before `query_options`.
+    #:
+    #: Set this on a base class to define base query options for its
+    #: subclasses, while still allowing those subclasses to define their own
+    #: additional query options via `query_options`.
+    #:
+    #: For example, set this to ``(raiseload('*', sql_only=True),)`` to prevent
+    #: all implicit SQL-emitting relationship loading, and force all
+    #: relationship loading to be explicitly defined via `query_options`.
+    base_query_options = ()
```
itajaja: is this actually needed couldn't you just refine query? would make for a cleaner api I guess

taion: i think this works better naming-wise. there are also a few variations – like raise, raise_on_sql, or... whatever

i think it'd be best to leave this as is for now, but revise later as we understand things better

PR #186, opened 2018-06-25
Cache view request args

taion:


flask_resty/authentication.py
```diff
@@ -2,17 +2,13 @@
 
 # -----------------------------------------------------------------------------
 
-REQUEST_CREDENTIALS_KEY = 'request_credentials'
```
itajaja: why remove this?

taion: idk, do you think having this random constant is that useful? i was looking at it more and thought it added more noise than anything else

taion:
@itajaja normally we just call `self.get_request_args()` with no kwargs set. i don't think we ever pass in any kwargs. since the kwargs themselves aren't nicely hashable, it seems like a good trade-off for performance and complexity to only handle the "normal" case.

itajaja:
fair enough. what about adding `request_args` property that calls into `get_request_args` then? I think it's clearer that's a memoized value

PR #189, opened 2018-06-27
feature: add get_raw_request_data

r1b:
This makes it easier to override the logic associated with getting the
request data w/o ovverriding the de-serializer.


flask_resty/view.py
```diff
@@ -87,14 +87,18 @@ def get_location(self, item):
         return flask.url_for(flask.request.endpoint, _method='GET', **id_dict)
 
     def get_request_data(self, **kwargs):
+        data_raw = self.get_raw_request_data()
+        return self.deserialize(data_raw, **kwargs)
+
+    def get_raw_request_data(self):
```
taion: can we rename this to `parse_request_data`?

PR #192, opened 2018-06-28
feature: support non-json mime types in testing

r1b:
* `get_body` no longer requires `application/json` - returns body as a
string for non-json mime types

* `assert_response` no longer requires `application/json`

Eventually one could imagine many plugins for parsing different mime
types but I think this is a natural next step.

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -80,20 +80,24 @@ def Shape(expected):
 
 
 def get_body(response):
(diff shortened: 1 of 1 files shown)


r1b:
This is also fine - this is how i have implemented it on my PR ;P

itajaja:
yeah I think that jimmy's proposal is fine :)

PR #194, opened 2018-07-02
feature: support custom json encoder

r1b:
This would save some typing in tests e.g when casting uuids.

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -11,6 +11,9 @@ from .utils import UNDEFINED
 
 
 class ApiClient(FlaskClient):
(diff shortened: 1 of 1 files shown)


taion:
ah, what's the motivation here?

r1b:
When sending request data that contains UUIDs I have to do `str(uuid)` or I get a JSONEncodeError

itajaja:
that's usually what we do (stringifying the uuid) /shrug. I agree it's a little bit daunting

taion:
I feel like I'd rather factor out parsers than add this to the view.

PR #195, opened 2018-07-09
Add skip_invalid to filtering

taion:


flask_resty/filtering.py
```diff
@@ -23,9 +23,10 @@ def filter_query(self, query, view, arg_value):
 
 
 class FieldFilterBase(ArgFilterBase):
-    def __init__(self, separator=',', allow_empty=False):
+    def __init__(self, separator=',', allow_empty=False, skip_invalid=False):
```
itajaja: where are you planning to use it?

taion: For id-like filters for use by data loaders. Even when they're actually UUIDs, IDs should semantically work like strings, so we should pretend that attempts to retrieve non-UUID-IDed fields get "nothing", rather than failing validation.

PR #202, opened 2018-07-18
Fix Python 3.7 compatibility

### LINKED ISSUE #200: Python 3.7 Compatibility
[Python 3.7 no longer allows trailing commas in generator expressions](https://bugs.python.org/issue32012) (see similar issue for [Django](https://code.djangoproject.com/ticket/28814)).

taion:
Fixes #200


taion:
gah that took a while

PR #219, opened 2018-11-02
Allow setting with_for_update in update

taion:


taion:
we decided against it as a default thing in https://github.com/4Catalyzer/flask-resty/pull/91 given that most update operations are idempotent

taion:
or idempotent is not the right term – "don't need locking" is more correct

itajaja:
oops sorry, wanted to approve, but I merged 🤦‍♂️ gotta go to sleep

taion:
same end result. thanks and have a good weekend!

PR #221, opened 2018-11-10
added get_deserializer to ColumnFilter

cancan101:
this allows subclasses to change how the deserializer is resolved

### DIFF
--- a/flask_resty/filtering.py
+++ b/flask_resty/filtering.py
@@ -140,9 +140,6 @@ class ColumnFilter(FieldFilterBase):
 
         self._column_name = arg_name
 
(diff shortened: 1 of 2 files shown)


itajaja:
what are you trying to achieve here? can't you just specify the specify deserializer in the subclass?

itajaja:
maybe you could achieve the same thing by overriding `get_field`?

cancan101:
@taion care to clarify?

taion:
maybe we should just add a `schema=Foo` kwarg to `ColumnFilter`

taion:
closing this out for now – the later proposals seem like a more targeted way to add this to the API

PR #223, opened 2018-11-15
Improve error reporting on status code mismatch

cancan101:
Straw man proposal for improving error reporting when there is a status mismatch in testing

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -96,6 +96,13 @@ def get_meta(response):
     return get_body(response)['meta']
 
 
+def get_status_code_mismatch_message(response):
+    if response.status_code >= 400:
+        return get_errors(response)
+    else:
+        return get_data(response)
+
+
 def assert_response(response, expected_status_code, expected_data=UNDEFINED):
     """Assert on the status and contents of a response.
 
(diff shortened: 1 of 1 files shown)


taion:
You use pytest, right? I wonder if we can hook into pytest's stuff to get "real" assertion support, like how it works with snapshottest.

cancan101:
```python
class Response:
    def __init__(self, expected_status_code, expected_data=UNDEFINED):
        self.expected_status_code = expected_status_code
        self.expected_data = expected_data

    def __eq__(self, response):
        if response.status_code != self.expected_status_code:
            return False

        elif self.expected_data is UNDEFINED:
            return True
        else:
            if 200 <= response.status_code < 300:
                response_data = get_data(response)
            else:
                response_data = get_errors(response)

            if not isinstance(self.expected_data, Predicate):
                expected_data = Shape(self.expected_data)

            return response_data == expected_data

def pytest_assertrepr_compare(config, op, left, right):
    import _pytest.assertion
    if isinstance(right, Response) and left.status_code != right.expected_status_code and left.status_code >= 400:
        return _pytest.assertion.pytest_assertrepr_compare(config, op, str(left.status_code), str(right.expected_status_code)) + [str(get_errors(left))]
    else:
        return _pytest.assertion.pytest_assertrepr_compare(config, op, left, right)
```
which gives something like:
```
>       assert response == Response(201, {"name": "Alex"})
E       AssertionError: assert '422' == '201'
E         - 422
E         + 201
E         [{'code': 'invalid_data', 'detail': 'xyz', 'source': {'pointer': '/data/name'}}]
```

taion:
i believe that between https://github.com/4Catalyzer/flask-resty/pull/239 and the "trap errors" flag, that this is no longer necessary

PR #224, opened 2018-11-19
Inspect violated constraints

taion:
cc @cancan101 @itajaja

This is obviously very rough and the final API will need to be more flexible, but the basic approach appears to work. Thoughts?

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -3,6 +3,7 @@ import itertools
 import flask
 from flask.views import MethodView
 from marshmallow import fields
+import sqlalchemy as sa
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Load
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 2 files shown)


flask_resty/view.py
```diff
@@ -415,20 +416,45 @@ def commit(self):
...
+                    column_names = ()
+        else:
+            column_names = ()
 
         flask.current_app.logger.exception("handled integrity error")
-        return ApiError(409, {'code': 'invalid_data.conflict'})
+
+        error = {'code': 'invalid_data.conflict'}
+        if column_names:
+            error['source'] = {'pointer': '/data/{}'.format(column_names[-1])}
```
taion: We'll actually have to go through the schema and figure out which deserializer field corresponds to this column.

taion: For multi-column unique constraints (e.g. tuple of `tenant_id`, `name`), we generally want to report the conflict on the last column, but this is something we need to tweak.

taion: We need to tweak the whole path thing, too. Generally the insert is only on the main model (i.e. no relationship inserts, most likely), but... eh.

cancan101: I think that assumption would require some level of documentation. Alternatively could make the default to show all columns and then have a hook in the view to allow the view to reduce what is shown.

taion: docs for sure. the issue is that we have a lot of unique constraints like `(tenant_id, name)`, and it would be odd to show an error on both fields in such cases

flask_resty/view.py
```diff
@@ -415,20 +416,45 @@ def commit(self):
...
+                # instead.
+                return error
+            else:
+                diag = original_error.diag
+
+                schema_name = diag.schema_name
+                table_name = diag.table_name
+                constraint_name = diag.constraint_name
+
+                insp = sa.inspect(self.session.bind)
```
itajaja: does it make more sense to do this at startup for each table, or to cache these?

flask_resty/view.py
```diff
@@ -415,20 +416,45 @@ def commit(self):
...
+                unique_constraints = insp.get_unique_constraints(
+                    table_name,
+                    schema=schema_name,
+                )
+
+                for unique_constraint in unique_constraints:
```
itajaja: ... so you can make this a map based on name. not sure if all this is prem optimization, but I think makes also the code cleaner. for with breaks smells to me

flask_resty/view.py
itajaja: we could also have a `pointers` field being an array of pointers. we are already not jsonapi compliant anyway :trollface:

cancan101:
I think one option would be to make this lookup be a set of utility function and then let the writer of the view decide how they want to do the mapping of the pg error to the api error.

taion:
@sloria since you looked at this – is there an easy way to go from attribute name to data key, given a schema?

sloria:
I think it'd be

```python
attr_to_data_key = {
    field.attribute or field_name: field.data_key or field_name 
    for field_name, field in schema.fields.items()
}
```

PR #226, opened 2018-12-19
feat: Allow returning item in update and delete

taion:


taion:
Note that this is non-breaking.

taion:
On second thought, better to make this breaking...

taion:
Eh, we can defer that for later.

flask_resty/view.py
```diff
@@ -477,18 +488,15 @@ def update(
...
-            return self.make_item_response(item)
-
-        return self.make_empty_response(item=item)
+        return self.make_updated_response(item, return_content=return_content)
 
     def destroy(self, id):
         item = self.get_item_or_404(id)
 
-        self.delete_item(item)
+        item = self.delete_item(item) or item
```
itajaja: this `or item` will never be hit because delete_item always returns an item, since there is an `or item` also there. so you either delete it from here or there. same for delete

taion: we don't want overriders of `delete_item` to have to return an item

taion:
The idea here is that actions that are semantically updates or deletes may not always correspond to updates or deletes in the database, e.g. with version models or with soft deletes, and that in those cases, we may want to return a new object for an update (for the update-with-version case), or have the option to still access the object after a delete (for a soft delete).

taion:
does this look gtg? @yodaiken

PR #233, opened 2019-02-12
[WIP] chore: Upgrade apispec

taion:
I thought I'd throw up this WIP. It's not working yet.

PR #234, opened 2019-02-12
chore: Pin apispec

taion:


taion:
@sloria this was screwing up the coverage ):

itajaja:
should we propose a motion in favor of dropping it? 🙄

taion:
let's discuss on #233

PR #239, opened 2019-03-01
feat: Mark test helpers as pytest plugin

taion:


taion:
why did this destroy my codecov coverage

tox.ini
```diff
@@ -6,16 +6,20 @@ passenv = DATABASE_URL
...
 extras =
     full: apispec,jwt
 
 commands =
     flake8 .
-    pytest {posargs:--cov}
+
+    # FIXME: We get screwed up coverage when using pytest-cov because we
+    #  declare ourselves as a pytest plugin.
+    coverage run -m pytest {posargs}
```
taion: sadface

taion: some weird interaction... i think pytest ends up rewriting the bytecode for everything for some reason. not worth spending more time here since this fix is functionally the same

itajaja:
nice!

PR #241, opened 2019-03-12
feat: Return response_data from assert_response

taion:


taion:
This makes the testing API slightly easier to use. Users now no longer need to explicitly call `get_data` or `get_errors` to get the response data in normal cases.

cancan101:
This only returns the response _if_ `expected_data` is provided which is a bit odd. ie the does not work:
```python
    data = assert_response(response, 200)
    xxx = data[0]['foo']
```
because this return line has not been changed:
```python
    if expected_data is UNDEFINED:
        return
```

taion:
@cancan101 I tried that with https://github.com/4Catalyzer/flask-resty/pull/242, but it breaks some stuff when the response isn't JSON (e.g. if the return code is 204). So far I've been passing in an explicit `ANY` as the third arg, but I agree that this isn't ideal. It'd be a breaking change in any event, though.

PR #242, opened 2019-03-12
fix: Always return response data from assert_response

taion:


cancan101:
perhaps return `None` in the case of 204 ?

taion:
I'm thinking `UNDEFINED` rather than `None`, perhaps.

taion:
cc @yodaiken @itajaja what do we think?

itajaja:
this makes sense. I think for 204 you shouldn't do anything special, but any other solution is fine

taion:
to be clear, you mean the current impl of returning `UNDEFINED` when the response data are empty?

itajaja:
correct, I mean, you shouldn't check `204` specifically

Issue #245, opened 2019-04-02
Failing tests


sfermigier:
On my box, some tests are failing, when launched either manually (`pytest`) or via `tox`.

```
flask-resty ❯ pytest --tb line
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.7.3, pytest-4.4.0, py-1.8.0, pluggy-0.9.0
rootdir: /Users/fermigier/git/flask/flask-resty
plugins: Flask-RESTy-0.20.10
collected 216 items

tests/test_args.py ..........                                                                                                                         [  4%]
tests/test_auth.py ...FF..F.........                                                                                                                  [ 12%]
tests/test_basic.py ...FF                                                                                                                             [ 14%]
tests/test_composite_id.py ...FF                                                                                                                      [ 17%]
tests/test_context.py ..                                                                                                                              [ 18%]
tests/test_decorators.py ....                                                                                                                         [ 19%]
tests/test_fields.py ........                                                                                                                         [ 23%]
tests/test_filtering.py .....................                                                                                                         [ 33%]
tests/test_jwt.py sssssssssssssssssssssssssssssssss                                                                                                   [ 48%]
tests/test_misc.py .........                                                                                                                          [ 52%]
tests/test_pagination.py .................................                                                                                            [ 68%]
tests/test_ping.py .                                                                                                                                  [ 68%]
tests/test_related.py .............                                                                                                                   [ 74%]
tests/test_sorting.py ......                                                                                                                          [ 77%]
tests/test_swagger.py ssssssssssssssssssss                                                                                                            [ 86%]
tests/test_testing.py .........                                                                                                                       [ 90%]
tests/test_utils.py .                                                                                                                                 [ 91%]
tests/test_view_errors.py .........xx.....F..                                                                                                         [100%]

========================================================================= FAILURES ==========================================================================
/Users/fermigier/git/flask/flask-resty/flask_resty/testing.py:87: AssertionError: assert 'text/html' == 'application/json'
/Users/fermigier/git/flask/flask-resty/flask_resty/testing.py:87: AssertionError: assert 'text/html' == 'application/json'
/Users/fermigier/git/flask/flask-resty/flask_resty/testing.py:87: AssertionError: assert 'text/html' == 'application/json'
/Users/fermigier/git/flask/flask-resty/flask_resty/testing.py:87: AssertionError: assert 'text/html' == 'application/json'
[...]

PR #246, opened 2019-04-02
fix: Fix testing utils with Werkzeug v0.15.x

taion:
Fixes https://github.com/4Catalyzer/flask-resty/issues/245

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -110,7 +110,7 @@ def assert_response(response, expected_status_code, expected_data=UNDEFINED):
     status_code = response.status_code
     assert status_code == expected_status_code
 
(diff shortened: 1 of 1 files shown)


flask_resty/testing.py
```diff
@@ -110,7 +110,7 @@ def assert_response(response, expected_status_code, expected_data=UNDEFINED):
     status_code = response.status_code
     assert status_code == expected_status_code
 
-    if response.content_length == 0:
+    if not response.content_length:
```
taion: this isn't 100% correct in all cases – content length of `None` could just mean it was omitted, but in practice this should be okay for the kinds of JSON requests that come out of flask resty

PR #247, opened 2019-04-03
chore: Add 'tests' extras

sloria:


setup.py
```diff
@@ -56,6 +56,14 @@ def run(self):
     extras_require={
         'apispec': ('apispec >= 0.39.0, < 0.40.0',),
         'jwt': ('PyJWT >= 1.4.0', 'cryptography >= 2.0.0'),
+        'tests': (
```
taion: isn't `tests_require` more correct here?

sloria: `tests_require` wouldn't really work with the current tox setup. Also, since the test command is `system('tox')`, we'd probably do `tests_require=('tox',)` if anything.

taion: i was trying to mimic e.g. flask in doing deps in tox so i could e.g. test against different dep version ranges, but i never got nearly that fancy

sloria: > is the goal here to make it easier to set up testing deps locally?

Yes

>  (even though nobody has any reason to do that)

Exactly.

> could we add e.g. a requirements-dev.txt?

Sure, but it's nice to use extras because you can easily combine them like in [attrs](https://github.com/python-attrs/attrs/blob/fdf053c3c5aaf5f47c92ee561fefda650b5f6fe0/setup.py#L38-L51). This is the setup we settled on in the marshmallow libs: https://github.com/marshmallow-code/marshmallow/issues/1020#issuecomment-433544037

PR #250, opened 2019-04-09
refactor: Remove apispec support

taion:
BREAKING CHANGE: Removes apispec support

This no longer works with the current apispec release, and we don't use it anyway.


taion:
cc @cancan101

flask_resty/api.py
```diff
@@ -171,12 +163,3 @@ def ping():
 class FlaskRestyState(object):
```
taion: we could remove this entirely and make the registered extension just be the `Api` object, but this is probably more future-proof

PR #252, opened 2019-04-23
feat: Handle custom body and error shapes in get_response

taion:


taion:
@itajaja I had to make another change. Take a quick look?

flask_resty/testing.py
```diff
@@ -100,7 +100,13 @@ def get_meta(response):
     return get_body(response)['meta']
 
 
-def assert_response(response, expected_status_code, expected_data=UNDEFINED):
+def assert_response(
+    response,
+    expected_status_code,
+    expected_data=UNDEFINED,
+    get_data=get_data,
```
itajaja: not sure if I am in favor of complicating this even more. if you have custom behavior, can't you do it outside of this function?

taion: the issue is that, if the shape doesn't match what's expected by the stock `get_data` or `get_error`, that those calls will just throw, so we'd end up having to reimplement this function in its entirety

there didn't seem to be a good way otherwise of getting this to always return data, but in an explicit way

PR #253, opened 2019-04-23
refactor: Simplify computing key for request_args

taion:


taion:
was trying to fix the weird coverage hit that came up here

Issue #255, opened 2019-05-03
Drop Python 2 support


sloria:
The time is near: https://pythonclock.org/

taion:
😮

hugovk:
Please see PR https://github.com/4Catalyzer/flask-resty/pull/257.

PR #257, opened 2019-05-22
chore: Drop support for legacy Python 2

hugovk:
Fixes https://github.com/4Catalyzer/flask-resty/issues/255.


.travis.yml
```diff
@@ -30,7 +28,3 @@ script:
 
 after_success:
   - codecov
-
-branches:
```
sloria: Why was this removed?

hugovk: It was removed so I could test my feature branch on Travis CI and be sure it was already passing, I didn't want to submit a broken PR to you! 

I can put it restore it if you prefer, although I would recommend leaving it in so others (and you!) can also test their feature branches too.

taion: I don't mind WIP PRs, actually. I added this specifically so I wouldn't see two Travis builds on my own PR branches.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist = 
-  py{27,35,37}-{base,full}-marshmallow2
+  py{35,37}-{base,full}-marshmallow2
```
taion:
```suggestion
  py{35,37}-{base,full}-marshmallow{2,3}
```

tox.ini
```diff
@@ -1,6 +1,6 @@
...
   py{35,37}-{base,full}-marshmallow3
```
taion:
```suggestion
```

tox.ini
taion: err, is this valid? @sloria

sloria: Yep. `tox -l` shows the correct output:

```
py35-base-marshmallow2
py35-base-marshmallow3
py35-full-marshmallow2
py35-full-marshmallow3
py37-base-marshmallow2
py37-base-marshmallow3
py37-full-marshmallow2
py37-full-marshmallow3
```

hugovk:
Updated!

taion:
Thanks so much!

hugovk:
You're welcome! Onwards with Python 3!

PR #262, opened 2019-05-28
fix: Fix Flask dependency range

taion:
I didn't realize v1.0.3 was cut. That explains why I had to make https://github.com/4Catalyzer/flask-resty/pull/261 to stay at 100% coverage.


taion:
I could test against v1.0.2 to keep coverage up, but there's no reason to do so.

PR #266, opened 2019-06-14
docs: add readthedocs config

sloria:
https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,11 @@
+version: 2
+sphinx:
+  configuration: docs/conf.py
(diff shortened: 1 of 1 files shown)


taion:
Is readthedocs still where Python docs generally go? We wouldn't put this on GH pages or something?

sloria:
Yeah, RTD is still the defactor standard for Python projects. Has some nice features that GH pages doesn't have, like the version switcher.

sloria:
https://flask-resty.readthedocs.io/en/latest/

PR #267, opened 2019-06-14
style: use pre-commit, fourmat, and friends

sloria:
Only includes config and files generated by fourmat for now, to make this easier to review.


taion:
I guess it's conventional to split up linting and testing?

sloria:
> I guess it's conventional to split up linting and testing?

Yeah, makes individual tox envs run faster.

sloria:
Bah, need to bump black version in fourmat to fix the trailing-comma feature detection.

Or maybe we should just drop py35 @taion ?

taion:
Our last Python 3.5 users are the people using the TensorFlow Docker image that still ships 3.5 by default. I don't think any of those images are using Flask-RESTy any more, though.

I'd be fine with dropping Python 3.5 support for now. If it causes problems, we can always bump versions and add it back.

PR #271, opened 2019-06-15
chore: use Travis for releasing

sloria:
Use Travis CI's rather undocumented build stages feature to automatically release to PyPI upon pushing a tag. Using build stages ensures that the release only happens if all other builds pass.

**ACTION REQUIRED**

@taion 

Need to add encrypted PyPI API token:

```
brew install travis
travis encrypt
... enter PyPI password then hit Ctrl+d...
```

Copy the value into the `secure` field.


setup.py
```diff
@@ -60,21 +60,8 @@ def run(self):
...
-    cmdclass={
-        "clean": system("rm -rf build dist *.egg-info"),
-        "docs": system("make -C docs html"),
-        "package": system("python setup.py sdist bdist_wheel"),
-        "publish": system("twine upload dist/*"),
-        "release": system("python setup.py clean package publish"),
-        "test": system("tox"),
+        "tests": ("coverage", "flake8", "psycopg2-binary", "pytest"),
     },
+    cmdclass={"test": system("tox")},
```
sloria: We can probably get rid of this and the `system` function altogether in favor of using `tox` directly. This is what we do in all the marshmallow libs (https://github.com/marshmallow-code/marshmallow/pull/1025).

I can do that in a follow-up PR.

sloria:
@taion Any interest in this?

If not, I can separate out the tox stuff into a different PR.

taion:
sorry i haven't had a chance to look in detail

sloria:
Oh ok no rush

taion:
Looking at this again – is it required to use user credentials on PyPI for this? It feels a little odd to give Travis my PyPI password.

sloria:
Looks like Travis has added support for API tokens recently. https://docs.travis-ci.com/user/deployment/pypi/ . Just change `user` to `__token__` and `password` to the token.

taion:
oops, i'm dumb. i revoked that token immediately and made a new one fwiw

taion:
and it looks like nobody published during the minute or two it was up, so i think we're okay

.travis.yml
```diff
@@ -15,9 +15,28 @@ jobs:
...
+      before_script: skip
+      script: skip
+      after_success: skip
+      deploy:
+        provider: pypi
+        user: "__token__"
+        password:
+          secure: "Z7mZWJfnuzmXH7j7ePbqZbexnSymAArUN1lQkqecguD6QkMwD8IIzAwM5EAld3pmrwdF6iIwu7utfZgaT8JoFavQYjcDLvvW+qvmK2i5K1YFcfrSpLoHifOFFkGmnVKpzLxDOO5KwDBgmbxWxUZB2AOmPUkDT1/Y0fg03f+sYXvSn7azT6FkhljRCP0+xgbqqZEs7nd/RVdCkdKHxsOlqaMC2CX/E64VBj0YAG0Fa9a8OCRmqbKmROnihqE8p0pKPAk7wj8VAJlaJmzNkSJDJLWs/thb1KG4CXcYWkcE/H7DWjvEXSPNCqPxsZ28WqcfKNmAYNMZBkvBW/99LpYd3zKM+ez/DyGRYzf8YOOQ7axIRwJOPExbZYZakny27hDcQEpaMLOuAP3NbMqaIHdJkQEDuTVHnDnFeCTn7rgHVzBaB7uQnaUD3z31EWDUOrN8jeUE3QnzR7FhhoWjrYkDwUFZpUVQzpzrTU17lTRLfJxXIUqqMBhdIgMZ/NGTLvXdDgdyPlt0iEPfg2VMXUl5JgwiKvWRK4raJY3y95TI8zHcUi/9XBM4YLPbQVYg/UuEJPt0p6A3YNnvHv855/SuZnn3Bmo+Vv5sI9DJZ/T7N86wWMDBdlEtboSXMc2Nv8UuO5kSxATmY+oCETh8f0J2zPqe83zrN1Es7cHYyccCLNA="
+        on:
+          tags: true
```
taion: @sloria is it not redundant to have both `if: tag IS present` _and_ `on: tags: true`?

taion:
This didn't seem to work... I'm going to try publishing v0.22.2 manually.

.travis.yml
sloria: Yeah, probably. The project I copied this config from transitioned from normal Travis PyPI deploys to Build stages, so likely that the `on: tags..` was just never removed.

sloria:
Hm, yeah, the stage executes but the actual release step doesn't happen. Strange. Will investigate when I have some extra time.

PR #276, opened 2019-06-25
feat: pull out HeaderAuthentication

r1b:
Test?

### DIFF
--- a/flask_resty/authentication.py
+++ b/flask_resty/authentication.py
@@ -88,9 +88,10 @@ class HeaderAuthentication(AuthenticationBase):
         authorization = flask.request.headers.get("Authorization")
 
         if authorization is None:
-            if self.credentials_arg is not None:
-                return flask.request.args.get(self.credentials_arg)
-            return None
+            if self.credentials_arg is None:
+                return None
+
+            return flask.request.args.get(self.credentials_arg)
 
         return self.get_credentials_from_authorization(authorization)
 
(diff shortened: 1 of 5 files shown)


flask_resty/authentication.py
```diff
@@ -62,3 +65,45 @@ class NoOpAuthentication:
...
+    #: used as credentials if the Authorization request header is missing.
+    credentials_arg = None
+
+    def get_request_credentials(self):
+        authorization = flask.request.headers.get("Authorization")
+
+        if authorization is None:
+            if self.credentials_arg is not None:
+                return flask.request.args.get(self.credentials_arg)
+            raise ApiError(401, {"code": "invalid_authorization.missing"})
```
taion: i think we just return None for “no credentials” here. the check for “credentials are required” is proper to authZ

flask_resty/authentication.py
```diff
@@ -62,3 +65,45 @@ class NoOpAuthentication:
...
+
+        return self.get_credentials_from_authorization(authorization)
+
+    def get_credentials_from_authorization(self, authorization):
+        if self.header_scheme is None:
```
taion: could just default this to Bearer. the other schemes all have handshakes and challenges that were not implementing

r1b:
Ping, I renamed this back to `HeaderAuthentication` since it can be instantiated as-is with a default scheme. Added tests, not sure why coverage is complaining.

r1b:
Ah nvm it was because I broke JWT :P

PR #279, opened 2019-07-09
feat: Extract HeaderAuthenticationBase

taion:


taion:
@r1b fyi - this lets us fix one of our callers that overrides the logic for getting the request token

taion:
@r1b fyi - this lets us fix one of our callers that overrides the logic for getting the request token

PR #283, opened 2019-07-11
 fix id check

### LINKED ISSUE #282: TypeError: argument of type 'Asset' is not iterable
Throws error when using POST method with code from example, with version 0.21.1, and `marshmallow==3.0.0rc8`

```
class AssetListView(AssetViewBase):
    def get(self):
        return self.list()

    def post(self):
        return self.create()
```
```
app\api\asset.py:21: in post
    return self.create()
..\xx-data\venv\lib\site-packages\flask_resty\view.py:1061: in create
    data_in = self.get_request_data(expected_id=expected_id)
..\xx-data\venv\lib\site-packages\flask_resty\view.py:245: in get_request_data
    return self.deserialize(data_raw, **kwargs)
..\xx-data\venv\lib\site-packages\flask_resty\view.py:709: in deserialize
    data = super().deserialize(data_raw, **kwargs)
..\xx-data\venv\lib\site-packages\flask_resty\view.py:291: in deserialize
    self.validate_request_id(data, expected_id)
```

```
self = <app.api.asset.AssetListView object at 0x0000020BFEAB3D68>, data = <Assert "hello" >, expected_id = False

    def validate_request_id(self, data, expected_id):
        """Check that the request data has the expected ID.
    
        This is generally used to assert that update operations include the
        correct item ID and that create operations do not include an ID.
    
        This works in one of three modes::
        - If `expected_id` is ``None``, do no checking
        - If `expected_id` is ``False``, check that no ID is provided
        - Otherwise, check that `data` has the expected ID
    
        :param data: The request data.
        :param expected_id: The ID or ID tuple, or False, or None.
        :raises ApiError: If the necessary IDs are not present and correct
        """
        if expected_id is None:
            return
    
        if expected_id is False:
            for id_field in self.id_fields:
>               if id_field in data:
E               TypeError: argument of type 'Asset' is not iterable
```

the `data` is an model object that loaded from `super().deserialize` method. So the `id_field` should be checked via `hasattr`.

JefferyLondon:
fix #282

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -372,9 +372,9 @@ class ApiView(MethodView):
         :return: The ID scalar or tuple.
         """
         if len(self.id_fields) == 1:
-            return data[self.id_fields[0]]
+            return getattr(data, self.id_fields[0])
 
-        return tuple(data[id_field] for id_field in self.id_fields)
+        return tuple(getattr(data, id_field) for id_field in self.id_fields)
 
     @request_cached_property
     def request_args(self):
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -351,7 +351,7 @@ class ApiView(MethodView):
 
         if expected_id is False:
             for id_field in self.id_fields:
-                if id_field in data:
+                if hasattr(data, id_field):
                     raise ApiError(403, {"code": "invalid_id.forbidden"})
             return
 


sloria:
Thanks for the PR! As mentioned on #282, the current behavior (schemas must deserialize to dicts) is by design.

Issue #275, opened 2019-06-16
Standardize query param list parsing


r1b:
Currently `request_args` are `foo=bar&foo=baz` where filters are `foo=bar,baz`.

r1b:
cc @stanley98yu

PR #286, opened 2019-07-17
fix: Support for comma-encoded query params

stanley98yu:
For: https://github.com/4Catalyzer/flask-resty/issues/275
On Asana: https://app.asana.com/0/779493871603507/1126826827214118

This PR allows for comma-encoded query params, having multiple params with the same name (i.e. `foo=bar&foo=baz`), or even a mixture of both.

If we want to only support comma-encoded query params, we can change `request_args` to just get the value for the key and split the param value, and not use any of the features of MultiDicts.

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -400,12 +400,9 @@ class ApiView(MethodView):
                 # KeyError for args that aren't present.
                 continue
 
-            if not isinstance(field, fields.List):
-                value = args.get(field_name)
-            else:
-                tmp, value = args.getlist(field_name), []
-                for v in tmp:
-                    value.extend(v.split(","))
+            value = args.get(field_name)
+            if isinstance(field, fields.List) or "," in value:
+                value = value.split(",") if value else []
 
             data_raw[field_name] = value
 
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -401,11 +401,12 @@ class ApiView(MethodView):
                 # KeyError for args that aren't present.
                 continue
 
-            value = args.get(field_name)
             if isinstance(field, fields.List) and not isinstance(
                 field, DelimitedList
             ):
                 value = args.getlist(field_name)
+            else:
+                value = args.get(field_name)
 
             data_raw[field_name] = value
 
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -402,13 +402,10 @@ class ApiView(MethodView):
                 continue
 
             value = args.get(field_name)
-            if isinstance(field, fields.List) and not hasattr(field, "delimiter"):
+            if isinstance(field, fields.List) and not isinstance(
+                field, DelimitedList
+            ):
                 value = args.getlist(field_name)
-            elif isinstance(field, DelimitedList):
-                try:
-                    value = value.split(field.delimiter)
-                except AttributeError:
-                    self.fail('invalid')
 
             data_raw[field_name] = value
 
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -400,9 +400,12 @@ class ApiView(MethodView):
                 # KeyError for args that aren't present.
                 continue
 
-            value = args.getlist(field_name)
-            if not isinstance(field, fields.List) and len(value) == 1:
-                value = value[0]
+            if not isinstance(field, fields.List):
+                value = args.get(field_name)
+            else:
+                tmp, value = args.getlist(field_name), []
+                for v in tmp:
+                    value.extend(v.split(","))
 
             data_raw[field_name] = value
 
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -14,6 +14,7 @@ from .authorization import NoOpAuthorization
 from .compat import MA2, schema_dump, schema_load
 from .decorators import request_cached_property
 from .exceptions import ApiError
+from .fields import DelimitedList
 from .utils import iter_validation_errors, settable_property
 
 # -----------------------------------------------------------------------------
@@ -401,8 +402,13 @@ class ApiView(MethodView):
                 continue
 
             value = args.get(field_name)
-            if isinstance(field, fields.List) or "," in value:
-                value = value.split(",") if value else []
+            if isinstance(field, fields.List) and not hasattr(field, "delimiter"):
+                value = args.getlist(field_name)
+            elif isinstance(field, DelimitedList):
+                try:
+                    value = value.split(field.delimiter)
+                except AttributeError:
+                    self.fail('invalid')
 
             data_raw[field_name] = value
 
--- a/flask_resty/fields.py
+++ b/flask_resty/fields.py
@@ -31,6 +31,7 @@ class RelatedItem(fields.Nested):
         # schema - in this context, they're actually not required.
         super(fields.Nested, self)._validate_missing(value)
 
+
 class DelimitedList(fields.List):
     """Same as `marshmallow.fields.List`, except can load from either a list or
     a delimited string (e.g. "foo,bar,baz"). Directly taken from webargs:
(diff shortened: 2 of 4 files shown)


stanley98yu:
This also really isn't `flask-resty`s fault since `flask.request.args` doesn't parse comma-encoded query params. I think this fix is sufficient, but if there's a better way to do the parsing, let me know.

flask_resty/view.py
```diff
@@ -400,9 +400,12 @@ def request_args(self):
                 # KeyError for args that aren't present.
                 continue
 
-            value = args.getlist(field_name)
-            if not isinstance(field, fields.List) and len(value) == 1:
-                value = value[0]
+            if not isinstance(field, fields.List):
```
taion: we probably shouldn't support both. perhaps for consistency, just do:

```python
value = args.get(field_name)

if isinstance(field, fields.List):
    # Treat empty strings as empty lists.
    value = value.split(",") if value else []
```

flask_resty/view.py
```diff
@@ -400,9 +400,12 @@ def request_args(self):
...
+                value = args.get(field_name)
+            else:
+                tmp, value = args.getlist(field_name), []
```
taion: the most pythonic way to do this is probably as a single list comprehension

```python
In [2]: a = ['a,b,c', 'd,e']

In [3]: [v2 for v1 in a for v2 in v1.split(',')]
Out[3]: ['a', 'b', 'c', 'd', 'e']
```

taion:
This should also get covered by a test in any event.

flask_resty/view.py
taion: cc @sloria

flask_resty/view.py
sloria: Or `chain`?

```python
In [1]: from itertools import chain

In [2]: a = ['a,b,c', 'd,e']

In [3]: a_lists = [each.split(',') for each in a]

In [4]: list(chain.from_iterable(a_lists))
Out[4]: ['a', 'b', 'c', 'd', 'e']
```

No strong preference here.

sloria: Or, to avoid the intermediate list:

```python

In [1]: from itertools import chain

In [2]: a = ['a,b,c', 'd,e']

In [3]: list(chain(*(each.split(',') for each in a)))
Out[3]: ['a', 'b', 'c', 'd', 'e']
```

which I think reads pretty nicely. Still no strong preference between Jimmy's and this approach. @stanley98yu takeaway here is that comprehensions are generally preferred over `for` loops with mutation.

flask_resty/view.py
taion: the question was more like – for list-like things in args, should we do comma-separated? or should we do multiple values?

or actually should we throw this whole thing out and just use webargs?

stanley98yu:
I added `"," in value` as another condition to check that fields not intended to be lists return a 422 status code if a comma is detected. Not sure if this is the best way, but it's succinct. Also, @taion should there also be a 422 if there are multiple query params with the same name? `flask.request.args.get` just gets the first param value, but doesn't report if there were more params with the same name.

taion:
I sorta think we don't want to block people from having commas in literal string values. Like, just in case they deal with it in application code.

That caveat with `.get()` is ubiquitous, so I'm okay with having it apply here as well.

flask_resty/view.py
sloria: Ha, probably not worth adding webargs just for this, but..webargs actually both. So webargs will call `args.getlist(...)` when parsing with a `List`. When you need to parse a comma-delimited list, you need to explicitly use [`DelimitedList`](https://github.com/marshmallow-code/webargs/blob/de061e037285fd08a42d73be95bc779f2a4e3c47/src/webargs/fields.py#L47) . We could yoink those bits here. wdyt, @taion ?

stanley98yu:
Could people use percent-encoding to differentiate between a comma in a literal string value and a comma as a delimiter?

flask_resty/view.py
stanley98yu: So if I have a field called `capabilities`, all I have to do is define it as a `DelimitedList` in my args schema to specify that I'll use commas as delimiters? And if I specify it as just `List`, I would use multiple param values?

sloria: Exactly. So what that would require here is

1. Adding `flask_resty.fields.DelimitedList`
2. Calling `getlist` when parsing with a `List` but not `DelimitedList`

```python
if isinstance(field, ma.fields.List) and not isinstance(DelimitedList):
    value = args.getlist(field_name)

# or duck type, like in webargs
if isinstance(field, ma.fields.List) and not hasattr(field, "delimiter"):
```

sloria:
> Could people use percent-encoding 

Seems reasonable to me.

flask_resty/view.py
taion: Can we do even simpler and define our own `DelimitedList` that doesn't extend `fields.List`, _only_ deals with delimited strings, and then just keep the code here the same?

taion: Wait, sorry, taking a step back. My question is – if you do `fields.List`, what's a reasonable thing for Flask-RESTy to do?

I don't recall why I decided to have it work with multiple args as opposed to comma-separated lists in the first place. It seems sort of inconsistent with all the `filtering`, &c. stuff.

It's might be weird that e.g. filtering operators only deal with comma-separated values, but it's definitely weird that the `request_args` scheme does something different and incompatible for lists.

sloria: I think I see what you mean. `filtering` and `sorting` transparently split on commas and pass those to the scalar marshmallow fields, whereas `request_args` expects duplicate args but does not handle comma-separated values. Sorta inconsistent.

I actually think this is fine behavior. `filtering` and `sorting` use a schema that defines the input and output shape for a model. `request_args`/`args_schema` is only dealing with input, so using `List` to handle duplicated query params is reasonable.

taion: okay, then i think the way to go is to add a `DelimitedList` field, then... but a slightly different one that only takes strings, i think

sloria: >  but a slightly different one that only takes strings, i think

To avoid the additional typecheck here? Sure, seems fine to me.

sloria: That said, I'm not sure there's a way to implement `DelimitedList` as a non-subclass of `List` without duplicating a bunch of `List`'s code. If the goal is to just avoid the typecheck here, it doesn't seem worth it.

sloria: I suppose you could use composition

```python
class DelimitedList(Field):
    def __init__(self, cls_or_instance, **kwargs):
       super().__init__(**kwargs)
       self._list_field = fields.List(cls_or_instance, **kwargs):

    def _deserialize(self, ....):
         try:
              value = value.split(self.delimiter)
          except AttributeError:
                self.fail('invalid')
          return self._list_field.deserialize(value, ...)
```

But that doesn't feel better than inheritance. wdyt, @taion ?

taion: I was just thinking for the sake of simplicity here. Otherwise won't the field get a single-element list with a comma-separated string, and then not be able to deal with it?

sloria: Not if we add the typecheck. To be clear, I'm proposing:

```python
            value = args.get(field_name)
            if isinstance(field, fields.List) and not hasattr(field, "delimiter"):
                value = args.getlist(field_name)
            data_raw[field_name] = value
```

flask_resty/view.py
```diff
@@ -400,9 +401,14 @@ def request_args(self):
                 # KeyError for args that aren't present.
                 continue
 
-            value = args.getlist(field_name)
-            if not isinstance(field, fields.List) and len(value) == 1:
-                value = value[0]
+            value = args.get(field_name)
+            if isinstance(field, fields.List) and not hasattr(field, "delimiter"):
```
taion: maybe something like:

```
if (
    (
        isinstance(field, fields.List)
        or not isinstance(field, fields.DelimitedList)
    )
    and len(value) == 1
):
    value = value[0]
```

don't inline the deserialization bits from here; `DelimitedList` should take care of it.

stanley98yu: Sorry, could you clarify how this works? I thought we had to return a list when the argument is a `List`. Also, what do you mean by "don't inline the deserialization bits"?

flask_resty/view.py
```diff
@@ -400,9 +401,11 @@ def request_args(self):
                 # KeyError for args that aren't present.
                 continue
 
-            value = args.getlist(field_name)
-            if not isinstance(field, fields.List) and len(value) == 1:
-                value = value[0]
+            value = args.get(field_name)
```
taion: put this in an `else` clause below? otherwise we're retrieving it for no reason

PR #248, opened 2019-04-06
feat: Support marshmallow 3

sloria:
close #210

PR #293, opened 2019-09-18
chore: drop support for marshmallow 2

sloria:
includes and closes #294


sloria:
Test failure is unrelated to this PR. Looks like the same error is happening on master

taion:
comparing with https://github.com/4Catalyzer/flask-resty/pull/248/files, do we want to do something w/the change for https://github.com/4Catalyzer/flask-resty/pull/248/files#diff-7014ed4582a13278460fe437b98a34ecR10 ?

otherwise this looks good... let's hold off for fixing CI before merging so we can see CI go green, though

sloria:
> do we want to do something w/the change for https://github.com/4Catalyzer/flask-resty/pull/248/files#diff-7014ed4582a13278460fe437b98a34ecR10 ?

I don't think so: the code as is is still correct

PR #294, opened 2019-09-18
fix: Fix example to match Marshmallow TZ semantics

taion:


docs/guide.rst
```diff
@@ -298,16 +298,16 @@ You can make requests using the `httpie <https://httpie.org/>`_ utility.
         "data": [
             {
                 "author_id": 2,
-                "created_at": "2019-06-16T01:09:33.450768+00:00",
+                "created_at": "2019-06-16T01:09:33.450768",
```
taion: Ick.

Maybe I should add a note that these are only naive because it's SQLite on the back end here?

taion: did that

taion:
@sloria boo i think you need to cherry-pick this to your branch actually. how does this look though?

taion:
@sloria can you cherry-pick this into #293, merge that, then close this?

sloria:
Yes will do

PR #295, opened 2019-09-20
fix: Default partial to None rather than False

taion:


taion:
never mind.

it would be nice if specifying `partial` on nested schemas worked

PR #296, opened 2019-10-04
chore: Try running CI against pypy

taion:


sloria:
I think for pure-python packages, pypy builds are mostly unnecessary and add maintenance burden.

FWIW I ended up removing pypy builds for my projects because 1) they increased CI build times, 2) installing pypy was a pain when I wanted to reproduce a build failure, and 3) when a pypy build failure did occur, it was never because of my code (usually an issue on Travis's end).

PR #297, opened 2019-10-23
[proposal] add immutable handling

itajaja:
just throwing it here. What do you guys think? this is non breaking, but the downside is that if someone overrides `update_item_raw`, they might lose this behavior. an alternative (probably better!) would be to change `deserialize` to accept a `skip_immutable_fields` flag

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -867,6 +867,10 @@ class ModelView(ApiView):
         :rtype: object
         """
         for key, value in data.items():
+            field = self.schema.fields.get(key)
+            if field and field.metadata.get('immutable'):
+                continue
+            
             setattr(item, key, value)
 
     def delete_item(self, item):


taion:
I like the idea of having immutable handling, but I don't like the implementation.

I think better would be to implement this on the schema level, such that when setting a `for_update` flag, we exclude those fields.

Then e.g. Marshmallow 3 can warn when they're specified when they shouldn't be, &c.

taion:
cc @sloria

sloria:
Yeah, would be nice to leverage marshmallow's existing `unknown` field functionality, but it might be tricky to do without being able to pass a schema class and constructor args within the view.

taion:
fine, we'll switch to schema classes :p

itajaja:
how would you use the unknown functionality?

sloria:
> how would you use the unknown functionality? 

The immutable fields could be excluded with `exclude`, then those fields will be treated as unknown fields per the [`unknown`](https://marshmallow.readthedocs.io/en/latest/quickstart.html#handling-unknown-fields) param.

itajaja:
I see. another problem (for any implementation so far proposed) is that it doesn't apply to nested schemas. maybe for rest semantics is fine, but it doesn't match closely what `dump_only` does.

PR #298, opened 2020-01-10
chore: Test against Python 3.8

taion:


sloria:
Ah, I was able to repro the failure after clearing the pre-commit cache. Will send a PR

taion:
Ah, thank you!

PR #299, opened 2020-01-10
fix: Handle check constraint violations as 409s

taion:


taion:
This is arguably a change, but it's more of a bugfix. A schema for e.g. a partial update can't always catch what would be a check constraint violation, so it makes sense to make those 409s rather than 500s.

PR #314, opened 2020-02-13
feat!: Add a proper upsert handler

taion:


flask_resty/view.py
```diff
@@ -189,23 +189,6 @@ def make_created_response(self, item):
             response.headers["Location"] = location
         return response
 
-    def make_updated_response(self, item, *, return_content=False):
```
taion: This was silly and pretty much always used incorrectly.

flask_resty/view.py
```diff
@@ -693,9 +676,9 @@ def get_item(
                 item = self.create_missing_item(id)
 
                 if will_update_item:
-                    # Bypass authorizating the save if we are getting the item
-                    # for update, as update_item will make that check.
-                    self.session.add(item)
+                    # Bypass authorizating the add if we are getting the item
+                    # to update it, as update_item will make that check.
+                    self.add_item_raw(item)
```
taion: might as well use this hook

itajaja:
> * Pretend that some implicit item exists, then run an actual update on it with the request data
> * Create a new item with the request data

isn't the latter a superset of the former? if you are doing PATCH, the semantics are clearly not well defined for upsert to start with, but the upsert route will work like what we already have. In other words I can't see an example that would require the first impl that the second wouldn't support, in which case we could drop with_for_update there, no?

taion:
there are some slight differences in semantics and for extensibility.

e.g. if for whatever reason the "create" fails, for (1), we give the user a 404. for (2), we more or less have to give the user a 403 or a 409 or whatever.

basically, the upsert bit in (2) doesn't run the "update" logic at all. so i don't know if that matters. it's a different code path and could lead to small differences for auth.

another way to think of it is – suppose there's no widget with ID `1` in the database. assume you want some sort of upsert if you do `PUT /widgets/1`.

what does `GET /widgets/1` return? if it returns a 404, then the logic per (2) is probably fine. if it returns a 200 with some default empty widget, then maybe you want to semantically go through update handlers...

sloria:
Seems sensible to me. So are you keeping the existing `update(create_missing)` behavior just for backwards compat, or do we expect there's a valid use case for it?

taion:
i mean, i'm really on the fence here. do you think it serves any purpose? or that anybody would _actually_ care and want it?

itajaja:
I am for smaller API surface

taion:
Okay, this is a proper breaking change now to update and missing item semantics. I've removed the old `create_missing` kwarg entirely, and now we have:

- For `retrieve` and `get_item`, replaced the `create_missing` option and the `create_missing_item` method with `create_transient_stub` and `create_stub_item`, which are more accurate names for what they do
- For `get_item`, remove `will_update_item`, which no longer is relevant
- For `update`, remove `create_missing`
- For `update`, remove `return_content`, and now always return a 200 with the content
- Add the new `upsert` handler

Thoughts?

flask_resty/view.py
```diff
@@ -1070,43 +1046,57 @@ def create(self, *, allow_client_id=False):
...
         """
-        # No need to authorize creating the missing item, as we will authorize
-        # before saving to database below.
-        item = self.get_item_or_404(
-            id,
-            with_for_update=with_for_update,
-            create_missing=create_missing,
-            will_update_item=True,
-        )
+        item = self.get_item_or_404(id, with_for_update=with_for_update,)
```
r1b: trailing comma

taion:
```suggestion
        item = self.get_item_or_404(id, with_for_update=with_for_update)
```

PR #316, opened 2020-02-15
refactor: Rename method for clarity

taion:


taion:
The key is not coming _from_ the JWT.

PR #318, opened 2020-02-26
Doc correct response for GenericModelView.destroy()

5uper5hoot:
Small doc fix: The docstring lists 201, but method returns 204 if successful.


taion:
Thanks!

PR #321, opened 2020-03-16
chore: Go back to manual publishing

taion:


taion:
given that the travis publishes don't seem to work atm

PR #322, opened 2020-03-16
chore: Unpin pre-commit

taion:


setup.py
```diff
@@ -36,7 +36,6 @@
     ),
     extras_require={
         "docs": ("sphinx", "pallets-sphinx-themes"),
-        "lint": ("pre-commit ~= 1.17"),
```
taion: this didn't seem to be used anywhere... and pre-commit installs everything it needs anyway, so i don't think we need anything more here

PR #323, opened 2020-03-16
chore: Try to fix release

taion:


taion:
I think Travis may link `after_success` and `deploy` somehow...

sloria:
I think this would result in `codecov` getting called during the deploy stage. IIRC the parent `before_*` and `after_*` get propagated down to child jobs, hence all the `skip`s.

Anyway, this is probably more trouble than it's worth, so +1 to switching back to manual releases for now.

taion:
i need https://travis-ci.community/t/does-skipping-everything-else-in-a-job-skip-deploy/6461 it seems

PR #325, opened 2020-03-16
fix: Make objects kw-only for flush

taion:


taion:
cc @cancan101

itajaja:
curious, what bug did this cause?

taion:
nothing in particular; i just forgot to do it

PR #327, opened 2020-03-26
feat: Add an upsert_item method

taion:


taion:
I don't think this is the right way to go. I think we might want more of a `update_item_some_more_raw` hook that both `update_item_raw` and `create_item_raw` call.

flask_resty/view.py
```diff
@@ -1082,19 +1108,17 @@ def upsert(self, id, *, with_for_update=False):
...
-        except NoResultFound:
-            item = self.create_and_add_item(data_in)
-            self.commit()
-
-            return self.make_created_response(item)
-        else:
-            item = self.update_item(item, data_in) or item
-            self.commit()
+        item = self.upsert_item(item, data_in) or item
+        updated = inspect(item).persistent
```
itajaja: this might not work if for some reason someone changes how the creation and upsert method work. what about upsert_item returns a tuple `(item: T, created: bool)`?

taion: I know. Awkward either way. It's why I'd be more inclined to add a new hook.

PR #333, opened 2020-04-30
refactor: Extract ApiError.format_validation_error

taion:


taion:
@sloria Maybe this? Passing in a lambda isn't the most Pythonic API. Maybe having a "validation error formatter" class that we compose in would be more Pythonic, but it'd also be quite a bit clunkier.

flask_resty/view.py
```diff
@@ -288,12 +284,12 @@ def deserializer(self):
         """
         return self.schema
 
-    def format_validation_error(self, error):
-        """Convert a marshmallow validation error to a serializable form.
+    def format_validation_error(self, message, path):
```
sloria, line 287: suggestion (non-blocking): add typing, since this is public API?

```suggestion
    def format_validation_error(self, message: str, path: typing.Sequence[str]):
```

flask_resty/pagination.py
```diff
@@ -398,7 +394,7 @@ def decode_value(self, value):
         value = base64.urlsafe_b64decode(value)
         return value.decode()
 
-    def format_validation_error(self, message):
+    def format_validation_error(self, message, path):
```
sloria, line 397: thought: should these be `staticmethod`s since they don't use `self`?

sloria: IIRC, pylint warns against methods that don't use self, suggesting they be staticmethods. But pylint is suuper pedantic ¯\\_(ツ)_/¯

taion: the reason why i've mostly avoided things that could be static or class methods is that they heavily tie our hands when subclassing and overriding... e.g. if they call into anything another method, then all those methods cannot be converted to instance methods on child classes, without breaking stuff

so i end up just not doing class methods at all

flask_resty/view.py
taion: let me just add some types to the docstring for now... i don't want to have just a single function with a type signature sitting on its own

PR #343, opened 2020-09-02
feat: add ModelSorting

jquense:
it's simple, but seems to work.

### DIFF
--- a/flask_resty/sorting.py
+++ b/flask_resty/sorting.py
@@ -142,7 +142,7 @@ class Sorting(FieldSortingBase):
     users can sort requests by `name` ascending and `date` descending by
     making a ``GET`` request to::
 
-        /api/post?sort=title,-content_length
+        /api/comments/?sort=title,-content_length
 
     :param str field_names: The fields available for sorting.
         Names should match a column on your View model.
--- a/flask_resty/sorting.py
+++ b/flask_resty/sorting.py
@@ -161,9 +161,8 @@ class Sorting(FieldSortingBase):
         names = frozenset(field_names)
         duplicates = keys.intersection(names)
 
-        assert not (
-            duplicates
-        ), f"Sort field(s): {', '.join(duplicates)} cannot be passed as field names and keyword arguments"
+    if duplicates:
+        raise ValueError(f"Sort field(s) cannot be passed as both positional and keyword arguments: {duplicates}")
 
         self._field_names = names.union(keys)
         self._field_sorters = {
--- a/flask_resty/sorting.py
+++ b/flask_resty/sorting.py
@@ -1,4 +1,5 @@
 import flask
+import sqlalchemy as sa
 
 from .exceptions import ApiError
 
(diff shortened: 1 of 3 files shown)


flask_resty/sorting.py
```diff
@@ -169,3 +170,43 @@ def get_request_field_orderings(self, view):
...
+    request parameter to determine the sort order. For example, users can sort requests
+    by `name` ascending and `date` descending by making a ``GET`` request to::
+
+        /api/posts?sort=author,-content_length
+
+    :param str default: If provided, specifies a default sort order when the
+        request does not specify an explicit sort order.
+
+    :param dict kwargs: A mapping from sort names to order_by expressions
+    """
```
jquense, line 196: IDK how to write this, it's RST?

flask_resty/sorting.py
```diff
@@ -169,3 +170,43 @@ def get_request_field_orderings(self, view):
...
+
+    def __init__(self, default=None, **kwargs):
+        super().__init__(*kwargs.keys(), default=default)
```
jquense, line 199: maybe this should also allow for field name positional args, to pass through to Sorting?

flask_resty/sorting.py
itajaja: @r1b 	:P

itajaja (commented):
nice! any reason to not have this in `Sorting` itself, so that you can do

```py
sorting = Sorting('foo', bar=Post.bar)
```

to limit api surface/proliferation of classes

jquense:
> nice! any reason to not have this in Sorting itself, so that you can do

oo I think that, I split it out b/c I didn't know if it was contentious but that does seem like a nice simplification

itajaja:
I think that it's not a breaking change, so should be fine

itajaja (approved):

flask_resty/sorting.py
```diff
@@ -128,36 +129,67 @@ def get_request_field_orderings(self, view):
...
-    def __init__(self, *field_names, default=None):
-        self._field_names = frozenset(field_names)
+    def __init__(self, *field_names, default=None, **kwargs):
+        keys = frozenset(kwargs.keys())
+        names = frozenset(field_names)
+        duplicates = keys.intersection(names)
+
+        assert not (
+            duplicates
+        ), f"Sort field(s): {', '.join(duplicates)} cannot be passed as field names and keyword arguments"
```
sloria, line 166: suggestion: `ValueError` should be raised for invalid arguments.

nitpick: Since this is a developer-facing error, I don't think there's a need to format it as a comma-delimited list.

```suggestion
    if duplicates:
        raise ValueError(f"Sort field(s) cannot be passed as both positional and keyword arguments: {duplicates}")
```

flask_resty/sorting.py
```diff
@@ -128,36 +129,67 @@ def get_request_field_orderings(self, view):
...
-    One or both of `name` or `date` can be formatted in the `sort_arg` request
-    parameter to determine the sort order. For example, users can sort requests
-    by `name` ascending and `date` descending by making a ``GET`` request to::
+    One or both of `title` or `content_length` can be formatted in the
+    `sort_arg` request parameter to determine the sort order. For example,
+    users can sort requests by `name` ascending and `date` descending by
+    making a ``GET`` request to::
 
-        /api/widgets?sort=name,-date
+        /api/post?sort=title,-content_length
```
sloria, line 145: nitpick: We typically use plural nouns, and `comments` is less overloaded than `posts`.

```suggestion
        /api/comments/?sort=title,-content_length
```

flask_resty/sorting.py
```diff
@@ -128,36 +129,67 @@ def get_request_field_orderings(self, view):
...
 
     :param str field_names: The fields available for sorting.
+        Names should match a column on your View model.
```
sloria, line 148: nitpick:

```suggestion
        Names should match a column on your View's ``model``.
```

tests/test_sorting.py
```diff
@@ -147,3 +207,8 @@ def test_error_empty(client):
         400,
         [{"code": "invalid_sort", "source": {"parameter": "sort"}}],
     )
+
+
+def test_duplicate_fields(client):
+    with pytest.raises(AssertionError):
```
sloria, line 213: nitpick: Together with my other suggestion:

```suggestion
    with pytest.raises(ValueError, match="Sort field(s) cannot be passed as both positional and keyword arguments"):
```

sloria (approved):

PR #344, opened 2020-09-25
feat: add shell command that auto-imports schemas and models

sloria:
Add an enhanced `flask shell` command that automatically imports SQLAlchemy models and marshmallow schemas.

Much of this is adapted from [flask-konch](https://github.com/sloria/flask-konch/blob/master/flask_konch/cli.py) and private code that we have in Butterfly.

Image 2020-09-25 at 4 33 41 PM

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,11 +33,15 @@ setup(
         "marshmallow>=3.0.0",
         "SQLAlchemy>=1.0.0",
         "Werkzeug>=0.11",
+        "konch>=4.0",
     ),
     extras_require={
         "docs": ("sphinx", "pallets-sphinx-themes"),
         "jwt": ("PyJWT>=1.4.0", "cryptography>=2.0.0"),
         "tests": ("coverage", "psycopg2-binary", "pytest"),
     },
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -33,11 +33,15 @@
         "marshmallow>=3.0.0",
         "SQLAlchemy>=1.0.0",
         "Werkzeug>=0.11",
+        "konch>=4.0",
```
itajaja, line 36: this will get installed into prod too, I guess it's ok. would be possible to make it an optional dep but possibly not worth it? or maybe it's not even common practice, you'd know better

itajaja (commented):
nice!

setup.py
sloria: I think it's actually useful to have in prod. When diagnosing issues, I often SSH, install ipython, then import the models.

I considered making this an optional dep, but extras are kinda a pain to deal with in our pip-tools setup. konch is a small (one module) dep without any transitive deps itself.

itajaja (approved):

Issue #329, opened 2020-04-02
Interested in any help with documentation?


peterschutt:
Now that you have gone 1.0 I assume the api should be stable enough to start filling holes in the documentation.

I'd be happy to assist.

taion:
Yup, we're definitely planning on keeping the API more stable.

To date, we've mostly been using this internally, so I'm sure there's a lot that we ought to explicitly document that we haven't. Are there any parts where _you_ would like to see the documentation expanded?

5uper5hoot:
Don't get me wrong, the API is well documented and the code is very clear to read, it's more the Guide section of the docs that still have a lot of TODOs in areas that I'm particularly interested in - especially the section about getting big. 

In an earlier issue that I raised, you mentioned that you were moving away from the `schemas.py`, `models.py`, `views.py` pattern, toward grouping objects together by resource. I've gravitated toward the same pattern myself, however the documentation recommends the former pattern. Whether or not you still believe the documented pattern is best for small projects starting out, either the recommended starting structure, or the "getting big" section of the docs could be expanded to detail the alternative "resource" based structure. Also, I'd love to hear the project's recommendations on how to integrate the library with the Flask application factory pattern.

One area where I thought I could expand the current docs is that I found that `ColumnFilter` automatically handles comma separated values as I was reading the code trying to work out how I'd implement a similar thing. I don't believe this is covered in the docs, except that the `separator` param is documented in `FieldFilterBase`. Perhaps a simple example of this in action in the `Filtering` guide would be beneficial.

Also, an easy fix, but I wanted to confirm that GenericModelView.update() now will not return a 204 in any case, as previously it was an option? The 204 is still documented, and I'm pretty sure that it is no longer possible? (This was actually how I realized that you'd released 1.0 as I was reading code on GitHub which was 1.0, but executing in 0.x and thought I was going crazy for a little while 🤣).

As I mentioned, I'd be very happy to put some time helping out here and will fit in with however you suggest is the way forward? For example, would you like separate issues for each doc addition? A project to track tasks? Etc.

taion:
Those all sound like fantastic ideas. What do you think of this?
- If there's something you ran into, where you wish it were documented, just make a PR, and we'll work to get it merged
- If there's something you ran into that you haven't been able to accomplish, just note it here – chances are we'll either have a recipe we can share (by adding it into our docs), or we can make a new issue for a feature request

peterschutt:
[...]

PR #350, opened 2020-10-15
prose doc for success/error response

peterschutt:
For #329 

I had a go at adding some prose docs for the success/error response types that the app returns by default.

I've added them into the `Views` section of the docs but happy to shift them to anywhere you'd like. Likewise, happy to make any changes to the content based on feedback.

I've tried to summarize all of the different error codes that the app might throw and provide a summary for each in a table. Perhaps that table could be broken up and the error codes documented within the section of the docs to which they apply, but I'll wait for your feedback before I tweak anything.

Thanks!

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -102,6 +102,71 @@ The concrete view classes simply call the appropriate CRUD methods from `Generic
 
     Unimplemented HTTP methods will return ``405 Method not allowed``.
 
+By default, responses that contain data are structured per the `JSON:API Top Level object <https://jsonapi.org/format/#document-top-level>`_ pattern.
+
+.. code-block:: json
+
+    {
+        "data": [
+            {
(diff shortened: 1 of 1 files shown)


itajaja (commented):
nice :) I was wondering, why not adding the prose in the error code themselves? [jsonapi has a specific `detail` field](https://jsonapi.org/format/#error-objects) for this. So eg

https://github.com/4Catalyzer/flask-resty/blob/ecf9adc33a2cb7b7d08dcb6eceff1497d4ddefe2/flask_resty/view.py#L942

can become

```py
return ApiError(
	409,
	{
		"code": "invalid_data.conflict",
		"detail": "Database integrity error.",
	},
)
```

and so on.

peterschutt:
Oh yes, that would make sense! I'll do that in a separate PR.

PR #353, opened 2021-01-06
build: switch to github workflows

itajaja:
adapted from https://github.com/python-attrs/attrs/blob/master/.github/workflows/main.yml


sloria:
Can we easily get automatic releases on tagged commits?

itajaja:
> Can we easily get automatic releases on tagged commits?

getting the test/coverage part right now. There is probably a gh action for that

.github/workflows/main.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    steps:
+      - uses: "actions/checkout@v2"
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install in dev mode"
+        run: "python -m pip install -e .[dev]"
+      - name: "Import package"
+        run: "python -c 'import attr; print(attr.__version__)'"
```
sloria, line 96:
```suggestion
        run: "python -c 'import flask_resty; print(flask_resty.__version__)'"
```

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
   lint
-  py{36,37,38}-{base,full}
+  py-{base,full}
```
sloria, line 4: question: do we not need the `[gh-actions]` section, per the docs? https://github.com/ymyzk/tox-gh-actions#tox-gh-actions-configuration. The attrs config also includes it: https://github.com/python-attrs/attrs/blob/72f94a3b03d5c2f8211f499c1c3fc72413d3fd4e/tox.ini#L11

.github/workflows/main.yml
```diff
@@ -0,0 +1,111 @@
...
+          python -VV
+          python -m site
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install --upgrade coverage[toml] virtualenv tox tox-gh-actions
+      - name: "Run tox targets for ${{ matrix.python-version }}"
+        run: |
+          set -xe
+          TOXENV=py-full python -m tox
+          TOXENV=lint python -m tox
+          TOXENV=docs python -m tox
```
sloria, line 39: suggestion (non-blocking): it'd be ideal if we could just run `lint` and `docs` on the latest Python version rather than all of them. No biggie though.

sloria (commented):
Looking good! Just some drive-bys

.github/workflows/main.yml
itajaja: lol :swea

.github/workflows/main.yml
itajaja: yeah. inside a job there is no way to "escape" the matrix, so it would have to be done in a separate job

tox.ini
itajaja: yeah. so, I don't really know tox much, but as far as I can understand, doesn't using something like github actions defy the purpose of tox? gh action already has the python matrix, so seems like they fulfill similar tasks...

.github/workflows/main.yml
itajaja: ... or I could use `if` to just make it run on latest python

tox.ini
sloria: tox is also useful for local development. I use it for `tox -e lint` as well as reproducing CI failures. Removing Python version mgmt outside of tox does lessen the benefit, though.

.github/workflows/main.yml
```diff
@@ -0,0 +1,111 @@
...
+        run: "ls -l dist"
+      - name: "Check wheel contents"
+        run: "check-wheel-contents dist/*.whl"
+      - name: "Check long_description"
+        run: "python -m twine check dist/*"
+
+  install-dev:
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
```
sloria, line 83: _Probably_ don't need to test against mac and windows

.github/workflows/main.yml
sloria: Your call. It would be nice to separate each of the tox envs to take advantage of the parallel jobs...iirc GH allows quite a few of them (10, maybe?).

sloria (approved):
Approving to unblock. I'm sure there's marginal improvements to make here, but this seems to be working as is and is obviously better than the status quo.

itajaja:
> Approving to unblock. I'm sure there's marginal improvements to make here, but this seems to be working as is and is obviously better than the status quo.

I'll figure out the tox thing first :)

tox.ini
c0state: Just curious--do you use `tox-pyenv` or something similar? ie: how do you have multiple clean Python installs locally to test locally that's close to CI? (or is that the wrong strategy)

sloria:
@itajaja Are you still working on this? Would be good to get CI working 😬

itajaja:
seems to work :) now I understand how that tox gh-actions integration is supposed to work :)

sloria (approved):

sloria:
thought: we should test 3.9 too, but let's get this merged first

PR #356, opened 2021-03-24
feat: return the updated item after update_item_raw

AaronBuxbaum:
This is more consistent, and compliant with the documentation

Screen Shot 2021-03-24 at 10 34 23 AM

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -843,6 +843,7 @@ class ModelView(ApiView):
         """
         for key, value in data.items():
             setattr(item, key, value)
+        return item
 
     def delete_item(self, item):
         """Delete an existing item.


sloria (approved):

itajaja:
@sloria do you know what's the deal with test_shell failing?

sloria:
Oy, I'm not sure..maybe an incompatibility with newer versions of SQLA on 3.6?

Sorry I'm booked up today so can't look into it closely.

As a stopgap you could just skip those tests on 3.6 and add a TODO ¯\_(ツ)_/¯

jquense:
its versions not just 3.6, that's just the fastest one to fail here

jquense:
it also seems to just be in the context of running via tox, running without that and they work fine

jquense:
this is way over my head

sloria:
Oh, I see. Looks like sqla 1.4 broke the shell command on all versions. Looking into this now.

sloria:
All fixed; merging

PR #357, opened 2021-03-24
fix: support boolean sorts when using RelayCursorPagination

jquense:


flask_resty/pagination.py
```diff
@@ -449,6 +449,13 @@ def get_filter_clause(self, column_cursors):
         )
 
         column, asc, value = column_cursors[-1]
+
+        # SQL Alchemy won't let you > or < a boolean, so we convert
+        # to an integer, the DB's seem to handle this just fine
+        if isinstance(value, bool):
+            column = sa.cast(column, sa.Integer)
+            value = 1 if value else 0
```
sloria, line 457: nitpick: This might not be necessary; bools are just ints under the hood, so I'd be somewhat surprised if `column > [True|False]` didn't just work.

If you do need an int, you can do

```suggestion
            value = int(value)
```

sloria (approved):

jquense:
@sloria do you know whats going on with CI here?

itajaja:
see https://github.com/4Catalyzer/flask-resty/pull/356#issuecomment-806011067

itajaja (approved):

sloria:
Merged the dep compatibility fixes into here. All green, so merging.

PR #361, opened 2021-03-25
build: fix publishing in GitHub Actions (take 2)

sloria:
Publish after building the package


itajaja (commented):
this looks good, but do you know what was wrong with previous set up? curious

PR #362, opened 2021-04-08
Add assert messages?

jquense:
easy win? I don't know honestly!

### DIFF
--- a/flask_resty/testing.py
+++ b/flask_resty/testing.py
@@ -64,7 +64,7 @@ def assert_shape(actual, expected):
             if value is not UNDEFINED:
                 assert (
                     key in actual
-                ), f"Expected key {key} not found in: {actual}"
+                ), f"expected key {key!r} not found in: {actual!r}"
 
                 assert_shape(actual[key], value)
             else:
@@ -74,7 +74,7 @@ def assert_shape(actual, expected):
     elif isinstance(expected, Sequence):
         assert isinstance(
             actual, Sequence
-        ), "Received a Sequence but actual value is not a Sequence"
+        ), f"{actual!r} is not a Sequence"
 
         actual_len = len(actual)
         expected_len = len(expected)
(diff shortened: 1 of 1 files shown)


jquense:
@sloria any thoughts?

flask_resty/testing.py
```diff
@@ -61,15 +61,30 @@ def assert_shape(actual, expected):
...
             else:
                 assert key not in actual
     elif isinstance(expected, (str, bytes)):
         assert expected == actual
     elif isinstance(expected, Sequence):
-        assert isinstance(actual, Sequence)
-        assert len(actual) == len(expected)
+        assert isinstance(
+            actual, Sequence
+        ), "Received a Sequence but actual value is not a Sequence"
```
sloria, line 76: suggestion: include the `actual` value here

```suggestion
        ), f"{actual!r} is not a Sequence"
```

flask_resty/testing.py
```diff
@@ -61,15 +61,30 @@ def assert_shape(actual, expected):
         # superset of the expected items, rather than that they match.
         for key, value in expected.items():
             if value is not UNDEFINED:
-                assert key in actual
+                assert (
+                    key in actual
+                ), f"Expected key {key} not found in: {actual}"
```
sloria, line 66: nitpick: keep casing consistent. also probably makes sense to use `reprs`.

```suggestion
                ), f"expected key {key!r} not found in: {actual!r}"
```

flask_resty/testing.py
```diff
@@ -136,6 +149,12 @@ def assert_response(
     else:
         response_data = response.data
 
+    status_code = response.status_code
+
+    assert (
+        status_code == expected_status_code
+    ), f"Expected status code to match for request data {response_data}"
```
sloria, line 156: suggestion:

```suggestion
    ), f"expected status code {expected_status_code!r}, got {status_code!r}"
```

sloria (approved):
Just some small suggestions, otherwise good to go

flask_resty/testing.py
c0state: Didn't know about `!r` !, but how come it's desirable? Without it, if the input was a string, it would call `str()`, which might be more desirable from what I understand?

sloria: `repr`s will generally be more detailed than `str` representations, and for strings, it's more readable to have the quotes

```
In [9]: actual = "number"

In [10]: print(f"{actual} is not a string")
number is not a string

In [11]: print(f"{actual!r} is not a string")
'number' is not a string

```

PR #364, opened 2021-08-05
feat: add backwards pagination

jquense:
POC for this since it wasn't so hard. It breaks the encapsulation of the base class a bit but it was sort of weird to begin.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,9 @@ Install dependencies:
 # optionally include [jwt] for the full test suite, but requires cryptography,
 #  and its dependent binary libraries
 pip install -e .[dev]
+
+# Or for zsh users
+pip install -e .\[dev\]
 ```
 
 ### Running tests
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -1,5 +1,5 @@
 import base64
-from typing import Any, List, Optional, Tuple
+from typing import Any, List, Tuple
 from flask_resty.view import ModelView
 from flask_resty.sorting import FieldOrderings, FieldSortingBase
 import flask
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -1,10 +1,11 @@
 import base64
-from typing import Any, List, Tuple
-from flask_resty.view import ModelView
-from flask_resty.sorting import FieldOrderings, FieldSortingBase
 import flask
 import sqlalchemy as sa
 from marshmallow import ValidationError
+from typing import Any, List, Tuple
+
+from flask_resty.sorting import FieldOrderings, FieldSortingBase
+from flask_resty.view import ModelView
 
 from . import meta
 from .exceptions import ApiError
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -25,6 +25,11 @@ class PaginationBase:
     logic.
     """
 
+    def adjust_sort_ordering(
+        self, view, field_orderings: FieldOrderings
+    ) -> FieldOrderings:
+        return field_orderings
+
     def get_page(self, query, view):
         """Restrict the specified query to a single page.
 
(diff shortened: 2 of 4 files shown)


flask_resty/pagination.py
```diff
@@ -537,16 +582,39 @@ class RelayCursorPagination(CursorPaginationBase):
...
+
+        cursor_in, reversed = self.get_request_cursor(view, field_orderings)
+
+        if reversed:
+            field_orderings = self.reverse_field_ordering(field_orderings)
+
+            query = query.order_by(None)  # clear any existing order_by
+            query = view.sorting.sort_query_by_fields(
+                query, view, field_orderings
+            )
```
jquense, line 600: aight this feels bad. Sorting via the `Sorting` class is done prior to `get_page` so there is already an `order_by` on the query, then we add some extra order-by's in based on `id` if they aren't already included. It feels unnecessary...the Pagination really needs to own the entire order_by. 

For the reverse case we actually need to remove the existing order by, reverse it and then add it back. I suspect that we should fully replace the query sort in all cases but not confident it won't break something. DRF seems to take this approach as well

CONTRIBUTING.md
```diff
@@ -17,6 +17,9 @@ Install dependencies:
 # optionally include [jwt] for the full test suite, but requires cryptography,
 #  and its dependent binary libraries
 pip install -e .[dev]
+
+# Or for zsh users
+pip install -e .\[dev\]
```
c0state, line 22: Hmm, I don't recall ever needing this. Maybe something else is causing this?

jquense:
@Will-Johns

flask_resty/pagination.py
```diff
@@ -335,7 +346,17 @@ def get_sorting_and_missing_field_orderings(self, view):
...
-    def get_request_cursor(self, view, field_orderings):
+    def reverse_field_ordering(
+        self,
+        field_orderings: FieldOrderings,
+    ) -> FieldOrderings:
+        inverted = []
+        for field, order in field_orderings:
+            inverted.append((field, not order))
+
+        return tuple(inverted)
```
itajaja, line 357: nit, you could use list comp?

flask_resty/pagination.py
itajaja: i am not sure why we need to change the order by? the order, whether you use before or after, shouldn't change...

jquense: it's reversed b/c i don't know a way of doing a reverse LIMIT, so to get the previous page, we reverse the order by and flip the WHERE so instead of:
```sql
SELECT * 
FROM foo
WHERE id > 4 
ORDER BY id
LIMIT 10
```
You get 
```sql
SELECT * 
FROM foo
WHERE id < 4 
ORDER BY DESC id
LIMIT 10
```

itajaja: of course. that is annoying.

itajaja: isn't probably better if it's the sorting that tries to be aware of the pagination, and in that case flip it? or alternatively sorting calling into some pagination method to manipulate the sorting before it's applied?

jquense: it's good question...maybe? Do you have an API in mind, I have a hard time knowing what is "best" in this library :P 

Isn't weird for encapsulation that these view properties call into each other all the time?

itajaja: a little. however pagination and sorting are coupled by nature, it's not extrinsic complexity.

I think something like

```py
        field_orderings = self.get_request_field_orderings(view)
        if view.pagination:
            field_orderings = view.pagination.adjust_sorting(field_orderings)
        return self.sort_query_by_fields(query, view, field_orderings)
```

or even:

```py
        field_orderings = self.get_request_field_orderings(view)
        if view.pagination and view.pagination.reverse_sorting:
            field_orderings = field_orderings[::-1]
        return self.sort_query_by_fields(query, view, field_orderings)
```

flask_resty/pagination.py
```diff
@@ -271,49 +294,27 @@ def ensure_query_sorting(self, query, view):
...
 
-        query = view.sorting.sort_query_by_fields(
-            query, view, missing_field_orderings
-        )
-        field_orderings = sorting_field_orderings + missing_field_orderings
-
-        return query, field_orderings
+        # ignore the passed in sort so that it's consistent
+        # with further calls in get_page
+        return self.get_field_orderings(view)
```
itajaja, line 305: yeah this is bizarre too, as pagination and sorting contend for who sorts... I guess it's ok. can't say it's clean. but it's a slight change of default behavior, even for non-backward requests. not sure if actually breaks anything

jquense: do we think it's worth making a backwards compatible version here? This seems like it should be fine? I think that if previously the sorts were different between pagination and sorting it would have been wrong right?

jquense:
can someone approve this :P

CONTRIBUTING.md
jquense: square brackets have special meaning in zsh most of the time you need to escape them if you want the literal

itajaja:
go for it!

PR #368, opened 2021-10-12
fix: support python 3.10

c0state:


sloria:
https://github.com/4Catalyzer/flask-resty/blob/master/RELEASING.md

PR #369, opened 2021-10-20
fix: fix compatibility with newer marshmallow versions

sloria:
Relevant [changelog](https://marshmallow.readthedocs.io/en/stable/changelog.html) items:

> Field.fail is deprecated. Use Field.make_error instead.

> The use of missing/default field parameters is deprecated and will be removed in marshmallow 4. load_default/dump_default should be used instead.


sloria:
note: coverage miss here is expected since we're not testing against multiple marshmallow versions. we should probably test against lowest version of all deps, but that's outside the scope of this PR

PR #372, opened 2021-11-17
refactor: move upsert logic up inheritance chain (wip)

gnottobfly:
* moves some of the logic from upsert into upsert_item in ModelView
* use upsert_item in GenericModelView

see https://butterflynetwork.atlassian.net/browse/SHARK-58

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -846,6 +846,26 @@ class ModelView(ApiView):
             setattr(item, key, value)
         return item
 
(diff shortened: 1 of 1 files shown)


gnottobfly:
I missed that on the create case it uses `return self.make_created_response(item)` vs `return self.make_item_response(item)` on the update case.

Not sure how the callee would be able to identify whether the item was updated or created.

PR #373, opened 2021-11-17
refactor: move upsert logic up inheritance chain

gnottobfly:
* moves some of the logic from upsert into upsert_item in ModelView
* use upsert_item in GenericModelView

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -1114,10 +1114,12 @@ class GenericModelView(ModelView):
         :rtype: :py:class:`flask.Response`
         """
         data_in = self.get_request_data(expected_id=id)
-        item, created = self.upsert_item(id, data_in, with_for_update)
         
+        item, created = self.upsert_item(id, data_in)
         self.commit()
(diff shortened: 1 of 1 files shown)


flask_resty/view.py
```diff
@@ -846,6 +846,29 @@ def update_item_raw(self, item, data):
             setattr(item, key, value)
         return item
 
+    def upsert_item(self, id, data, with_for_update=False):
+        """Updates an existing item with the matching id or if the item does not exist yet, create and insert it.
```
sloria, line 850: nitpick: per PEP8, use imperative in docstrings

```suggestion
        """Update an existing item with the matching id or if the item does not exist yet, create and insert it.
```

sloria (approved):

flask_resty/view.py
sloria: just gonna make this change cuz it's literally one character

sloria (approved):

sloria:
this is released in 1.10.0

PR #374, opened 2021-12-08
feat: add page info to connection meta

jquense:
Absolutely not sure this is the right way to do this. Basically we need info about totals and cursor index in order to provide paging UI that looks like 
image

I've no sense on the most efficient way to return this info tho. should we remove the selected columns from the query? is that safe? is it ok to rely on what SQL Alchemy generates here? should we provide hooks to customize it?

### DIFF
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -564,17 +564,51 @@ class RelayCursorPagination(CursorPaginationBase):
     https://facebook.github.io/relay/graphql/connections.htm.
     """
 
+    page_info_arg = "page_info"
+
+    def __init__(self, *args, default_include_page_info=False, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._default_include_page_info = default_include_page_info
+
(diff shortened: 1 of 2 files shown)


sloria (approved):

flask_resty/pagination.py
```diff
@@ -564,25 +564,61 @@ class RelayCursorPagination(CursorPaginationBase):
...
     """
 
+    page_info_arg = "page_info"
+
+    def __init__(self, *args, default_include_page_info=False, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._default_include_page_info = default_include_page_info
+
+    def get_page_info(self, query, view, field_orderings, cursor):
+        include = flask.request.args.get(
```
sloria, line 574: thought (non-blocking): would be nice to properly deserialize the string to bool with marshmallow if it's easy. nbd tho

sloria: maybe

```python
    include = self.deserialize_value(fields.Boolean(), flask.request.args.get(....)
```

PR #376, opened 2022-01-03
feat: add shell_setup

manon-pilaud:


flask_resty/shell.py
```diff
@@ -163,6 +169,7 @@ def context_formatter(full_context: dict):
             ],
         )
     )
+
```
manon-pilaud, line 172:
```suggestion
```

sloria (approved):
Nice!

thought (non-blocking): for symmetry it probably makes sense to have `RESTY_SHELL_TEARDOWN`, but that can wait until we actually need it.

PR #381, opened 2022-02-07
chore: update pyjwt

c0state:


flask_resty/jwt.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, jwk_set, *args, **kwargs):
         self.jwk_set = jwk_set
 
     def decode(self, jwt, **kwargs):
-        unverified_header = self.get_unverified_header(jwt)
+        unverified_header = jwt_lib.get_unverified_header(jwt)
```
c0state, line 75: This got moved to a new `PyJWS` object which can be accessed directly via the `jwt` module

sloria (approved):
thanks for completing a 3-point ticket for us 😁 

Firefox 2022-02-07 at 15 01 28

c0state:
Can someone also add @matt-mclaughlin-quantum-si as cloud team member also? Thanks.

sloria:
invite sent to @matt-mclaughlin-quantum-si

c0state:
This should be a v3 release correct because https://github.com/4Catalyzer/flask-resty/compare/v2.0.0...master#diff-be86804fe2ca36af9c77b587cb3b41811668560f76cef3acdf9730b5e6f25798R51 could be fundamentally different? If so, I'll push a `v3.0.0` tag?

c0state:
> hm, i wouldn't normally count dependency upgrades as a breaking change so long as there's no public-facing API change, so I think this can be 2.1.0

https://github.com/4Catalyzer/flask-resty/compare/v2.0.0...master#diff-be86804fe2ca36af9c77b587cb3b41811668560f76cef3acdf9730b5e6f25798R51 returns the jwt module though which could have some breaking changes I think?

sloria:
oh i see, i missed that. i suppose that's technically public API, yes. otoh, we at bfly aren't using `.pyjwt` anywhere. also, it'd be reasonable to say that only _documented_ API should be considered public.

sloria:
but go ahead and bump up to v3 to be conservative

PR #383, opened 2022-03-15
fix: paginating with nulls by RelayCursorPaginator

d-wysocki:
RelayCursorPaginator wasn't able to paginate and sort by nullable rows.
1. `None` values in cursor was deserializing into  string`"None"` and query looks like `col_name > "None"`
2. Add `IS NULL` filter clause for columns that are nullable, example `col_name > aTest OR col_name IS NULL`


d-wysocki:
Solution will be in other PR

PR #384, opened 2022-03-31
fix: get the schema deserialized view args if available

c0state:
Does this make sense? It seems like the args_schema deserialized view args should be used, if available

### DIFF
--- a/flask_resty/filtering.py
+++ b/flask_resty/filtering.py
@@ -351,7 +351,7 @@ class Filtering:
         :return: The filtered query
         :rtype: :py:class:`sqlalchemy.orm.query.Query`
         """
-        args = flask.request.args
+        args = view.request_args if view.args_schema else flask.request.args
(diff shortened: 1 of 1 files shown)


flask_resty/filtering.py
```diff
@@ -351,7 +351,7 @@ def filter_query(self, query, view):
         :return: The filtered query
         :rtype: :py:class:`sqlalchemy.orm.query.Query`
         """
-        args = flask.request.args
+        args = view.request_args if view.args_schema else flask.request.args
```
c0state, line 354: Reason being that caller might have some custom manipulation of the query args that could then get picked up by the filters

sloria:
Makes sense to me. Can you add a test please?

c0state:
Actually closing, what we needed was `load_default`. Not sure if this PR is 100% correct

PR #388, opened 2022-04-06
chore: update to version 4.0.1 with route test fixes

c0state:


flask_resty/__init__.py
```diff
@@ -44,4 +44,4 @@
     pass
 
 
-__version__ = "4.0.0"
+__version__ = "4.0.1"
```
c0state, line 47: patch update should be OK for this diff I think?
https://github.com/4Catalyzer/flask-resty/compare/v4.0.0...master

sloria (approved):

PR #389, opened 2022-04-11
fix: paginating with nulls by RelayCursorPaginator [PUFF-246]

d-wysocki:
This PR fixing issues:
1. Support paginating by NULLS in `RelayCursorPaginator`
2. Make consistent sorting by NULLS (nulls last and nulls first)
3. Deserialize "None" into None

### DIFF
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -483,7 +483,7 @@ class CursorPaginationBase(LimitPagination):
         value += (3 - ((len(value) + 3) % 4)) * b"="  # Add back padding.
         value = base64.urlsafe_b64decode(value).decode()
 
-        return None if value == "None" else value
+        return None if value == str(sa.null()) else value
 
     def deserialize_value(self, field, value):
         if value is None:
@@ -639,6 +639,8 @@ class CursorPaginationBase(LimitPagination):
         return ".".join(self.encode_value(value) for value in cursor)
 
     def encode_value(self, value):
+        if value is None:
+            value = sa.null()
         value = str(value)
         value = value.encode()
         value = base64.urlsafe_b64encode(value)
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -537,7 +537,9 @@ class CursorPaginationBase(LimitPagination):
         # For an empty and_() construct, use and_(True, *args).
         previous_clauses = sa.and_(True)
         for column, _, value in column_cursors:
-            previous_clauses = sa.and_(previous_clauses, column.isnot_distinct_from(value))
+            previous_clauses = sa.and_(
+                previous_clauses, column.isnot_distinct_from(value)
+            )
         return previous_clauses
 
     @staticmethod
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -486,7 +486,7 @@ class CursorPaginationBase(LimitPagination):
         return None if value == "None" else value
 
     def deserialize_value(self, field, value):
-        if not field.required and value is None:
+        if value is None:
             return None
         return (
             field.deserialize(value)
@@ -535,12 +535,9 @@ class CursorPaginationBase(LimitPagination):
     def get_previous_clause(column_cursors):
         # Invoking and_() without arguments is deprecated, and will be disallowed in a future release.
         # For an empty and_() construct, use and_(True, *args).
-        previous_clauses = sa.and_()
+        previous_clauses = sa.and_(True)
         for column, _, value in column_cursors:
-            if value is None:
-                previous_clauses = sa.and_(previous_clauses, column.is_(None))
-            else:
-                previous_clauses = sa.and_(previous_clauses, column == value)
+            previous_clauses = sa.and_(previous_clauses, column.isnot_distinct_from(value))
         return previous_clauses
 
     @staticmethod
@@ -580,8 +577,7 @@ class CursorPaginationBase(LimitPagination):
         column, asc, value = column_cursors[-1]
 
         current_clause = self._prepare_current_clause(column, asc, value)
-        if not str(previous_clauses):
-            return current_clause
+
         return sa.and_(previous_clauses, current_clause)
 
     def make_cursors(self, items, view, field_orderings):
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -16,6 +16,9 @@ from .utils import if_none
 # -----------------------------------------------------------------------------
 
 
+_NULL = str(sa.null())
+
+
 class PaginationBase:
     """The base class for pagination components.
 
@@ -483,7 +486,7 @@ class CursorPaginationBase(LimitPagination):
         value += (3 - ((len(value) + 3) % 4)) * b"="  # Add back padding.
         value = base64.urlsafe_b64decode(value).decode()
 
-        return None if value == str(sa.null()) else value
+        return None if value == _NULL else value
 
     def deserialize_value(self, field, value):
         if value is None:
@@ -640,7 +643,7 @@ class CursorPaginationBase(LimitPagination):
 
     def encode_value(self, value):
         if value is None:
-            value = sa.null()
+            value = _NULL
         value = str(value)
         value = value.encode()
         value = base64.urlsafe_b64encode(value)
--- a/flask_resty/pagination.py
+++ b/flask_resty/pagination.py
@@ -533,14 +533,14 @@ class CursorPaginationBase(LimitPagination):
 
     @staticmethod
     def get_previous_clause(column_cursors):
-        # Invoking and_() without arguments is deprecated, and will be disallowed in a future release.
-        # For an empty and_() construct, use and_(True, *args).
-        previous_clauses = sa.and_(True)
-        for column, _, value in column_cursors:
-            previous_clauses = sa.and_(
-                previous_clauses, column.isnot_distinct_from(value)
-            )
-        return previous_clauses
+        if not column_cursors:
+            return None
+        clauses = [
+            column.isnot_distinct_from(value)
+            for column, _, value in column_cursors
+        ]
+
+        return sa.and_(*clauses)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


tests/test_pagination.py
```diff
@@ -331,6 +348,62 @@ def test_relay_cursor_sorted(client, data):
     }
 
 
+def test_relay_cursor_sorted_none_desc(client, data_with_nulls):
```
jquense, line 351: I know we don't generally have any tests  like this, but it would (imo) be good to have a few tests around the expected SQL output, both to ensure that it's correct, and to prevent subtle regressions

flask_resty/pagination.py
```diff
@@ -528,12 +531,27 @@ def get_filter(
...
+    @staticmethod
+    def get_previous_clause(column_cursors):
+        # Invoking and_() without arguments is deprecated, and will be disallowed in a future release.
+        # For an empty and_() construct, use and_(True, *args).
+        previous_clauses = sa.and_()
+        for column, _, value in column_cursors:
+            if value is None:
+                previous_clauses = sa.and_(previous_clauses, column.is_(None))
+            else:
+                previous_clauses = sa.and_(previous_clauses, column == value)
```
jquense, line 543:
```suggestion
   previous_clauses = sa.and_(previous_clauses, column.isnot_distinct_from(value))
```
maybe?

flask_resty/pagination.py
```diff
@@ -528,12 +531,27 @@ def get_filter(
...
+        return previous_clauses
```
jquense, line 544: does this produce something different compared to:

```py
return sa.and_(column.isnot_distinct_from(value) for column, _, value in column_cursors)
```

might avoid the empty and usage?

flask_resty/pagination.py
```diff
@@ -528,12 +531,27 @@ def get_filter(
...
+    @staticmethod
+    def _handle_nullable(column, value, is_nullable):
+        if is_nullable:
+            return sa.or_(column.is_(None), (column > value))
+        else:
+            return column > value
 
-        column, asc, value = column_cursors[-1]
+    def _prepare_current_clause(self, column, asc, value):
+        is_nullable = getattr(column.expression, "nullable", True)
```
jquense, line 554: is this not `column.nullable`? what is `expression` here?

flask_resty/pagination.py
```diff
@@ -542,10 +560,28 @@ def get_filter_clause(self, column_cursors):
             value = int(value)
 
         if asc:
-            current_clause = column > value
+            if value is None:
+                return None
+            elif value is not None:
+                current_clause = self._handle_nullable(
+                    column, value, is_nullable
+                )
```
jquense, line 568: the assumption here is that the column is always `nulls first` for ascending, right?

flask_resty/pagination.py
```diff
@@ -542,10 +560,28 @@ def get_filter_clause(self, column_cursors):
...
+            else:
+                current_clause = column > value
         else:
-            current_clause = column < value
+            if value is None:
+                current_clause = column.isnot(None)
```
jquense, line 573: why does the descending case need different logic for null handling?

tests/test_pagination.py
```diff
@@ -331,6 +348,62 @@ def test_relay_cursor_sorted(client, data):
...
+    response = client.get("/relay_cursor_widgets?sort=-name")
+    assert_response(
+        response, 200, [{"id": "8", "name": None}, {"id": "7", "name": None}]
```
jquense, line 354: i know this the default behavior with psql, but i find this sort behavior so odd, and am i little concerned our users won't understand it. e.g will they expect that "no exam type" is sorted last when sorting alphabetically. @sloria do you have any opinions on whether our users will expect this?

flask_resty/sorting.py
```diff
@@ -115,7 +115,7 @@ def get_criteria(self, view, field_orderings):
     def get_criterion(self, view, field_ordering):
         field_name, asc = field_ordering
         column = self.get_column(view, field_name)
-        return column if asc else column.desc()
+        return sa.nullslast(column) if asc else sa.nullsfirst(column.desc())
```
jquense, line 118: is the explicit nulls * to be safe and consistent or does this change something from the default?

flask_resty/pagination.py
```diff
@@ -542,10 +560,28 @@ def get_filter_clause(self, column_cursors):
...
+            else:
+                current_clause = column < value
+        return current_clause
+
+    def get_filter_clause(self, column_cursors):
+        previous_clauses = self.get_previous_clause(column_cursors[:-1])
+        column, asc, value = column_cursors[-1]
 
+        current_clause = self._prepare_current_clause(column, asc, value)
+        if not str(previous_clauses):
```
jquense, line 583: what is this checking?

flask_resty/pagination.py
d-wysocki: For DESC NULLS came first, so we need to check if there are any objects that have any value example
```
SELECT * FROM widgets
WHERE widget.name IS NULL AND widget.id > 3 OR widget.name IS NOT NULL

```

flask_resty/sorting.py
d-wysocki: Different  databases have different behavior related to how NULLs are sorted (for example sqlite (used for tests in flask-resty) and postgres handle nulls order by in different way). So we need to make it consistent to avoid issues reletad to this.

flask_resty/pagination.py
d-wysocki: After adding `isnot_distinct_from` this line is unnecessary

flask_resty/pagination.py
d-wysocki: good catch, thanks

flask_resty/pagination.py
d-wysocki: good catch, thanks

flask_resty/pagination.py
jquense: sorry, ok I misspoke asc -> nulls last. So for non NULL values we have to assume there may be upcoming null values, hence the handling

flask_resty/pagination.py
d-wysocki: expression is for `column_propety` its also works for regular fields

tests/test_pagination.py
```diff
@@ -331,6 +348,62 @@ def test_relay_cursor_sorted(client, data):
...
+    )
+    cursor = get_meta(response)["cursors"][-1]
+    response = client.get(f"/relay_cursor_widgets?sort=name&cursor={cursor}")
+    assert_response(
+        response,
+        200,
+        [{"id": "1", "name": "Whatzit"}, {"id": "4", "name": None}],
+    )
+    cursor = get_meta(response)["cursors"][-1]
+    response = client.get(f"/relay_cursor_widgets?sort=name&cursor={cursor}")
```
jquense, line 401: can we also test with backwards pagination? eg. the `before` arg. Shouldn't cause any issues, but worth covering

tests/test_pagination.py
jquense: we should probably not ask and act surprised later i think

flask_resty/pagination.py
```diff
@@ -481,10 +481,13 @@ def decode_cursor(self, cursor: str) -> Tuple[str, ...]:
     def decode_value(self, value: str):
         value = value.encode("ascii")
         value += (3 - ((len(value) + 3) % 4)) * b"="  # Add back padding.
-        value = base64.urlsafe_b64decode(value)
-        return value.decode()
+        value = base64.urlsafe_b64decode(value).decode()
+
+        return None if value == "None" else value
```
sloria, line 486: it's a bit odd to string-match against `None`; it might make sense to encode `None` as `null` or `NULL` in `encode_value`. wdyt?

jquense: I like that

jquense: I think the idea would be to decode it to `None` or `sql.null()` still. But serialize to the as "null" for the cursor.

I think we don't need to do it in this PR tho

jquense (approved):
Approving to unblock. Looks great, we may want to doxument the follow ups in a ticket

flask_resty/pagination.py
akjanik: I prefer to work with python data types whenever possible, mixing with sql.null can be confusing, I never know if I should use one or another. But  it's SQLA thing.
BTW can we change to something like
```return value if value is None else base64.urlsafe_b64decode(value).decode()```
so we don't need to compare to `"None"` (it looks odd)? Or is there a some hidden middle step that prevent this?

Comment (jquense) [reply flask_resty/pagination.py:486]: some value needs to be encoded in the cursor or it won't work correct. `null` is the actual value here, `None` is simply the python representation of it, hence it makes sense to serialize the None as `'null'` in the same way None serialized to the DB `null` in a query

Comment (jquense) [reply flask_resty/pagination.py:486]: which is to say this decode_value function is decoding sql column values that were stored in a b64'd string

File: flask_resty/pagination.py [hunk 12]
```diff
@@ -483,7 +483,7 @@ def decode_value(self, value: str):
         value += (3 - ((len(value) + 3) % 4)) * b"="  # Add back padding.
         value = base64.urlsafe_b64decode(value).decode()
 
-        return None if value == "None" else value
+        return None if value == str(sa.null()) else value
```
sloria, line 486: nitpick (non-blocking): can probably just use `"NULL"` here or make a constant.

```python
_NULL = str(sa.null())

def decode_value(..):
   return ... if value == _NULL else value
   
def encode_value(..):
  if value is None:
    value = _NULL
  else:
    value = str(value)
```

this saves a bit of type-casting (which normally wouldn't matter but this is a rather "hot" code path, as the encoding/decoding happens for every item in the queryset)

sloria (approved):

flask_resty/pagination.py
d-wysocki: done

PR #391, opened 2023-05-30
fix: order of add item authorization check seems incorrect

c0state:
The order of the call to authorize whether an item can be added seems unusual. Opening for discussion.

### DIFF
--- a/flask_resty/view.py
+++ b/flask_resty/view.py
@@ -784,8 +784,8 @@ class ModelView(ApiView):
 
         :param object item: The item to add.
         """
-        self.add_item_raw(item)
         self.authorization.authorize_save_item(item)
+        self.add_item_raw(item)
 
     def add_item_raw(self, item):
         """As with `add_item`, but without the authorization check.


flask_resty/view.py
```diff
@@ -784,8 +784,8 @@ def add_item(self, item):
 
         :param object item: The item to add.
         """
-        self.add_item_raw(item)
         self.authorization.authorize_save_item(item)
```
c0state, line 787: Shouldn't we do the auth check first? In practice not too big an issue as the session doesn't commit and rolls back, but this seems odd...

sloria: my guess is that the intention is to authorize just before calling db.session.commit().

have you run into the use case where this call order matters? 👀

c0state: > have you run into the use case where this call order matters? 👀

Only noticed it while doing somewhat unrelated Sqlalchemy 2 upgrade work on our end. I don't think it matters in practice, but it does mean that it adds the model into the session even if the caller is unauthorized temporarily

sloria: yeah, i agree that the current behavior is a bit odd. i've no objection to merging this

c0state:
Was going to fix the tests up first, but going to merge for now as they're already broken on mainline CI

PR #392, opened 2023-09-29
fix: updates for latest flask sqlalchemy

c0state:


flask_resty/view.py
```diff
@@ -493,7 +493,7 @@ class ModelView(ApiView):
     @settable_property
     def session(self):
         """Convenience property for the current SQLAlchemy session."""
-        return flask.current_app.extensions["sqlalchemy"].db.session
+        return flask.current_app.extensions["sqlalchemy"].session
```
shatch-bfly, line 496: suggestion: could be made backwards compatible (support both styles)

shatch-bfly:
question: does this need to be a major release? seems like everything breaking can be made backwards compatible? most of the churn seems to be isolated to `tests`?

flask_resty/routing.py
```diff
@@ -1,21 +0,0 @@
...
-try:
-    from werkzeug.routing import RequestPath
-except ImportError:  # pragma: no cover
-    # werkzeug<1.0
-    from werkzeug.routing import RequestSlash as RequestPath
-
-# -----------------------------------------------------------------------------
-
-
-class StrictRule(Rule):
```
c0state, line 12: Decided to remove this as this doesn't work with latest and we don't use it at all on our end.

flask_resty/utils.py
```diff
@@ -52,9 +53,9 @@ def __init__(self, get_default):
     def __get__(self, instance, owner):
         if instance is None:
             return self
-        try:
+        if hasattr(instance, self.internal_field_name):
```
c0state, line 56: Slightly more explicit rewrite

flask_resty/view.py
c0state: Don't disagree in principle, but there's a bit more changes in tests, etc which make this a bit annoying. 

Upstream dependencies also are major versions, so I figure we could also bump in sync. 

Lastly I don't think we have much external usage, so I opted to optimize for 4c umbrella org usage.

c0state:
This should be mostly ready I think but need to see what that Python 3.7 is coming from...

sloria:
i updated the required checks to remove the 3.7 build. should be good to go!

flask_resty/routing.py
shatch-bfly: I don't see any references here either

flask_resty/utils.py
shatch-bfly: no preference, is there any perf penalty (like dbl lookup?)

shatch-bfly (approved):

flask_resty/utils.py
c0state: Hmm, didn't consider but I suspect the former was actually less performant (caught exception vs key lookup)

.pre-commit-config.yaml
```diff
@@ -3,13 +3,14 @@ repos:
   rev: v2.31.1
   hooks:
     - id: pyupgrade
-      args: [--py37-plus]
+      args: [--py39-plus]
 - repo: https://github.com/4Catalyzer/fourmat
   rev: v0.11.1
   hooks:
     - id: fourmat
+      additional_dependencies: [setuptools>69]
```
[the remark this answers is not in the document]
c0state: I think it was related to removal of setuptools by default in Python 3.12
https://github.com/python/cpython/issues/95299
https://python-poetry.org/blog/announcing-poetry-1.7.0/#official-support-for-python-312

Although, I'm not sure it's needed here now. Might have been something in my local env as I have 3.12 as the default.

PR #394, opened 2024-09-05
chore: update supported python versions and refine ci workflow

c0state:


tox.ini
```diff
@@ -1,10 +1,11 @@
 [tox]
-envlist = py39-{base,full}, py310-{base,full}, docs, lint
+envlist = py310-{base,full}, py311-{base,full}, py312-{base,full}, docs, lint
 
 [gh-actions]
 python =
-    3.9: py39-base, py39-full
-    3.10: py310-base, py310-full, lint, docs
+    3.10: py310
```
c0state, line 6: `tox-gh-actions` should run all envs of a given prefix:
https://github.com/ymyzk/tox-gh-actions#factor-conditional-settings-python-version

matt-mclaughlin-quantum-si (approved):
@c0state looks good, although dropping support for a python version may require a major version

c0state:
> @c0state looks good, although dropping support for a python version may require a major version

Hmm, maybe should add back:
https://devguide.python.org/versions/

3.9 still has a year left. Thought it was EOL or already EOL...

PR #401, opened 2025-04-25
fix: use feature detection for marshmallow version

c0state:
Refine marshmallow dependency version check logic due to upstream deprecation

Deprecated in version https://github.com/marshmallow-code/marshmallow/blob/5429f0d4c6e346dc751599073cab67d66eefbcb2/CHANGELOG.rst#3210-2024-02-26


sloria (commented):
Better to use `importlib.metadata.version("marshmallow")`, which will work on all versions

Also I'd recommend dropping support for marshmallow<3.13.0; it's old 😛

flask_resty/filtering.py
```diff
@@ -12,10 +13,7 @@
 # -----------------------------------------------------------------------------
 
 # Field.missing is deprecated in favor of Field.load_default in marshmallow 3.13.0
-try:
-    _USE_LOAD_DEFAULT = marshmallow.__version_info__ >= (3, 13)
-except AttributeError:
-    _USE_LOAD_DEFAULT = True
+_USE_LOAD_DEFAULT = importlib.metadata.version("marshmallow")
```
sloria: this doesn't looks quite right. still need to check the version number here

c0state: 🤦🏼 should be fixed now. Sorry for the delay.

flask_resty/filtering.py
```diff
@@ -12,7 +14,10 @@
 # -----------------------------------------------------------------------------
 
 # Field.missing is deprecated in favor of Field.load_default in marshmallow 3.13.0
-_USE_LOAD_DEFAULT = marshmallow.__version_info__ >= (3, 13)
+_MARSHMALLOW_VERSION = importlib.metadata.version("marshmallow")
+_USE_LOAD_DEFAULT = packaging.version.parse(
```
c0state: Using `packaging.version` for semantically checking version. *Should* be OK I think? Note that this handles https://peps.python.org/pep-0440/ vs full semver I believe...

sloria: i usually use `packaging.version.Version`, but this probably is equivalent

sloria (approved):

flask_resty/filtering.py
sloria: oh also, i think you'll need to add `packaging` to setup_requires in setup.py

sloria (changes requested):
oh, i jsut saw you bumped up the minimum version of marshmallow in setup.py. in that case, the version checking is no longer necessary, right?

c0state:
> oh, i jsut saw you bumped up the minimum version of marshmallow in setup.py. in that case, the version checking is no longer necessary, right?

Ah right, further refined in latest

sloria (approved):

